PR #10, opened 2022-09-30
Make collection of non-existent files more clear

pyrco:
(DIS-962)

### DIFF
--- a/acquire/collector.py
+++ b/acquire/collector.py
@@ -9,7 +9,12 @@ from itertools import groupby
 from pathlib import Path
 from typing import Iterable, List, Union, Optional, Set, Type, Sequence, Any
 
-from dissect.target.exceptions import FileNotFoundError
+from dissect.target.exceptions import (
+    FileNotFoundError,
+    NotADirectoryError,
+    NotASymlinkError,
+    SymlinkRecursionError,
+)
 from dissect.target.helpers import fsutil
 
 from acquire.utils import get_formatted_exception, StrEnum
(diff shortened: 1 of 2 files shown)


acquire/collector.py
```diff
@@ -341,6 +341,9 @@ def collect_path(
             path = self.target.fs.path(path)
 
         try:
+            # If a path does not exist, is_dir(), is_file() and is_symlink() will return False (and not raise an
+            # exception), so we need to explicitly check for this.
+            exists = path.exists()
```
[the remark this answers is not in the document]
pyrco: I keep `self.target.fs.lexists(path)` as there are some issues with `path.get().lexists()` and the underlying `FileEntry.get()` being implemented inconsistently. Some implementations do resolve and others do not resolve symlinks.

pyrco:
@Schamper  Note: I added calls to reporting when a symlink or an unknown file type are collected (they are now reported as a failed collection).

acquire/collector.py
```diff
@@ -341,6 +341,9 @@ def collect_path(
             path = self.target.fs.path(path)
 
         try:
+            # If a path does not exist, is_dir(), is_file() and is_symlink() will return False (and not raise an
+            # exception), so we need to explicitly check for this.
+            exists = self.target.fs.lexists(str(path))
```
Schamper, line 346: I think there's one issue here currently, where if `path` is already a `TargetPath` but from a different filesystem. That might be an issue here: https://github.com/fox-it/acquire/blob/main/acquire/acquire.py#L256

I think you said that `path.get().lexists()` unfortunately doesn't work reliably, so then I guess it unfortunately has to become something like `path._fs.lexists(str(path))`?

pyrco: Hmm, that won't work then. I'll temporarily use the `path._fs.lexists(str(path))` way then until the issues with `TargetPath.get()` and `FileEntry.lexists()` are resolved.

Schamper (dismissed):

Schamper (approved):

PR #14, opened 2022-10-27
Remove esxi memory limit using vsish

Miauwkeru:
It changes the memory limits of the group running acquire with vsish.
Each of these commands get logged inside the log file.
Once acquire is done executing, or another error occurrs, it will attempt to restore the changed values.
Vsish will raise an error if the esxi group uses too much memory.
As the vsish commands get logged, you can retry the last command in the log file.

(DIS-881)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,3 +9,6 @@ long_description_content_type = text/markdown
 license_files = LICENSE, COPYRIGHT
 classifiers =
   Programming Language :: Python :: 3
+
+[options]
+python_requires = ~=3.9


acquire/esxi.py
```diff
@@ -0,0 +1,85 @@
...
+        self.mem_scheme: Dict[str, str] = dict()
+
+    def setup(self):
+        """"""
+        self.group_id = self._get_group_id()
+        self.mem_scheme = self._get_memory_scheme()
+        self._set_memory_limits(max="unlimited", min_limit="unlimited")
+
+    def reset(self):
+        """Put memory limits back"""
```
Schamper, line 31:
```suggestion
        """Put memory limits back to their original values."""
```

acquire/esxi.py
```diff
@@ -0,0 +1,85 @@
...
+            [
+                "set",
+                f"/sched/groups/{self.group_id}/memAllocationInMB",
+                f"max={max}",
+                f"minLimit={min_limit}",
+            ]
+        )
+
+    def _get_memory_scheme(self):
+        """Retrieve and parse the current memory scheme from an ESXI host."""
```
Schamper, line 74:
```suggestion
        """Retrieve and parse the current memory scheme from an ESXi host."""
```

Schamper (changes requested):

Schamper (approved):

PR #19, opened 2022-11-11
Proposal to improve error handling

cecinestpasunepipe:
Upon failure to open a target, this update checks whether the user is admin/root and displays a hint.

(DIS-1648)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1734,7 +1734,7 @@ def main():
         log.info(target)
     except Exception:
         if not is_user_admin():
-            log.error("Hint: you might try to re-run as administrator/root.")
+            log.error("Failed to load target, try re-running as administrator/root.")
             exit(1)
         log.exception("Failed to load target")
         raise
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -1733,6 +1733,9 @@ def main():
         target = Target.open(target_path)
         log.info(target)
     except Exception:
+        if not is_user_admin():
+            log.error("Hint: you might try to re-run as administrator/root.")
```
Schamper, line 1737:
```suggestion
            log.error("Failed to load target, try re-running as administrator/root.")
```
?

acquire/acquire.py
```diff
@@ -1733,6 +1733,9 @@ def main():
         target = Target.open(target_path)
         log.info(target)
     except Exception:
+        if not is_user_admin():
+            log.error("Failed to load target, try re-running as administrator/root.")
+            exit(1)
```
Schamper, line 1738:
```suggestion
            parser.exit(1)
```
I don't think it actually matters, just for consistency sake because it's what's used a couple lines above.

Schamper (approved):

PR #25, opened 2022-11-22
Add open handles plugin

Poeloe:
(DIS-728)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -21,8 +21,6 @@ from dissect.target.plugins.os.windows import iis
 from dissect.target.plugins.os.windows.log import evt, evtx
 
 from acquire.collector import Collector, get_full_formatted_report, get_report_summary
-from acquire.dynamic.windows.collect import collect_open_handles
-from acquire.dynamic.windows.handles import serialize_handles_into_csv
 from acquire.dynamic.windows.types import HandleType
 from acquire.esxi import esxi_memory_context_manager
 from acquire.hashes import (
@@ -1310,6 +1308,13 @@ class OpenHandles(Module):
 
     @classmethod
     def run(cls, target: Target, cli_args: dict[str, any], collector: Collector):
+        if not sys.platform == "win32":
+            log.error("Open Handles plugin can only run on Windows systems! Skipping...")
+            return
+
+        from acquire.dynamic.windows.collect import collect_open_handles
+        from acquire.dynamic.windows.handles import serialize_handles_into_csv
+
         log.info("*** Acquiring open handles")
 
         handle_types = cli_args.handle_types
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -14,15 +14,16 @@ from collections import defaultdict
 from pathlib import Path
 
 from dissect.target import Target, exceptions
-from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.filesystems import ntfs
 from dissect.target.helpers import fsutil
+from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.plugins.os.windows import iis
 from dissect.target.plugins.os.windows.log import evt, evtx
 
 from acquire.collector import Collector, get_full_formatted_report, get_report_summary
 from acquire.dynamic.windows.collect import collect_open_handles
 from acquire.dynamic.windows.handles import serialize_handles_into_csv
+from acquire.dynamic.windows.types import HandleType
 from acquire.esxi import esxi_memory_context_manager
 from acquire.hashes import (
     HashFunc,
@@ -1295,7 +1296,14 @@ class FileHashes(Module):
 
 
 @register_module("--handles")
-@module_arg("--file-handle-type", action="append", help="Only collect file handles")
+@module_arg(
+    "--handle-types",
+    action="extend",
+    help="Collect only specified handle types",
+    type=HandleType,
+    choices=[h.value for h in HandleType],
+    nargs="*",
+)
 @local_module
 class OpenHandles(Module):
     DESC = "Open handles"
@@ -1304,17 +1312,18 @@ class OpenHandles(Module):
     def run(cls, target: Target, cli_args: dict[str, any], collector: Collector):
         log.info("*** Acquiring open handles")
 
-        file_only = cli_args.file
+        handle_types = cli_args.handle_types
 
         collector.bind(cls)
         try:
-            handles = collect_open_handles(["file-handle-type"] if file_only else None)
+            handles = collect_open_handles(handle_types)
             csv_compressed_handles = serialize_handles_into_csv(handles)
 
             collector.write_bytes(
                 f"{collector.base}/{collector.METADATA_BASE}/open_handles.csv.gz",
                 csv_compressed_handles,
             )
+            log.info("Collecting open handles is done.")
         finally:
             collector.unbind()
 
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -21,7 +21,7 @@ from dissect.target.plugins.os.windows import iis
 from dissect.target.plugins.os.windows.log import evt, evtx
 
 from acquire.collector import Collector, get_full_formatted_report, get_report_summary
-from acquire.dynamic.windows.types import HandleType
+from acquire.dynamic.windows.named_objects import NamedObjectType
 from acquire.esxi import esxi_memory_context_manager
 from acquire.hashes import (
     HashFunc,
@@ -1298,8 +1298,8 @@ class FileHashes(Module):
     "--handle-types",
     action="extend",
     help="Collect only specified handle types",
-    type=HandleType,
-    choices=[h.value for h in HandleType],
+    type=NamedObjectType,
+    choices=[h.value for h in NamedObjectType],
     nargs="*",
 )
 @local_module
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -21,6 +21,8 @@ from dissect.target.plugins.os.windows import iis
 from dissect.target.plugins.os.windows.log import evt, evtx
 
 from acquire.collector import Collector, get_full_formatted_report, get_report_summary
+from acquire.dynamic.windows.collect import collect_open_handles
+from acquire.dynamic.windows.handles import serialize_handles_into_csv
 from acquire.esxi import esxi_memory_context_manager
 from acquire.hashes import (
     HashFunc,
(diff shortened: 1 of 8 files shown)


acquire/acquire.py
```diff
@@ -1292,6 +1294,31 @@ def get_specs(cls, cli_args):
...
+@module_arg("--file-handle-type", action="append", help="Only collect file handles")
+@local_module
+class OpenHandles(Module):
+    DESC = "Open handles"
+
+    @classmethod
+    def run(cls, target: Target, cli_args: dict[str, any], collector: Collector):
+        log.info("*** Acquiring open handles")
+
+        file_only = cli_args.file
```
Schamper, line 1307: I'm confused by these arguments. You specify `--file-handle-type` as argument for the module, with action append, but check for the `file` argument, and treat it like a boolean?

Poeloe: Yep, forgot about that one... Was in doubting how to tackle it. Fixed it now :)

acquire/dynamic/windows/types.py
```diff
@@ -1,12 +1,180 @@
...
+    USHORT,
+    WCHAR,
+)
+from enum import IntEnum
 
 PVOID = ctypes.c_void_p
 NTSTATUS = ULONG
 NULL = None
 
+ULONG_PTR = ctypes.c_size_t
```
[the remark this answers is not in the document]
Poeloe: Resulted in some issues, so kept it separate for now

tests/test_acquire_command.py
```diff
@@ -4,7 +4,8 @@
 
 import pytest
 
-from acquire.acquire import (
+patch('ctypes.windll', create=True).start() # noqa
+from acquire.acquire import ( # noqa
```
pyrco, line 8: Are these `# noqa`s needed? I guess because the `patch()` is between imports? Is it not possible to patch after the imports?

If not maybe better to only disable the specific warning/error.

tests/test_acquire_command.py
```diff
@@ -4,7 +4,8 @@
 
 import pytest
 
-from acquire.acquire import (
+patch('ctypes.windll', create=True).start() # noqa
```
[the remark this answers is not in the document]
Poeloe: Indeed, I guess this does not work because of the fact that it is Windows only

tests/test_acquire_command.py
Poeloe: Made it more specific, thanks :)

acquire/dynamic/windows/ntdll.py
```diff
@@ -29,6 +31,17 @@
...
+ntdll.NtQueryInformationFile.argtypes = (
+    HANDLE,
+    ctypes.POINTER(IO_STATUS_BLOCK),
+    LPVOID,
+    ULONG,
+    DWORD,
+)
+ntdll.NtQueryInformationFile.restype = NTSTATUS
+ntdll.NtQuerySystemInformation.restype = NTSTATUS
+ntdll.NtQueryObject.restype = NTSTATUS
```
pyrco, line 43: Shouldn't these also have `argtypes` defined?

acquire/dynamic/windows/ntdll.py
```diff
@@ -29,6 +31,17 @@
 
 
 ntdll = ctypes.windll.ntdll
+ntdll.NtQueryInformationFile.argtypes = (
```
pyrco, line 34: I get the idea of not putting `NtQueryInformationFile` in this module's namespace, but for the sake of consistency it maybe should, like `NtOpenDirectoryObject` and others below.

acquire/dynamic/windows/ntdll.py
Poeloe: Done!

acquire/dynamic/windows/types.py
```diff
@@ -1,12 +1,201 @@
...
+)
+from enum import Enum, IntEnum
 
 PVOID = ctypes.c_void_p
 NTSTATUS = ULONG
 NULL = None
 
+ULONG_PTR = ctypes.c_size_t
+
+DUPLICATE_SAME_ACCESS = 0x00000002
```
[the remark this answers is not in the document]
Poeloe: Placed it in ProcessAccess IntEnum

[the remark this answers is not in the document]
pyrco: Would it be an issue that it has the same value as `PROCESS_CREATE_THREAD` in that same enum?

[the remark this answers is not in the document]
Poeloe: Well, in this case it does not, since here the enum is used only as a one-way lookup (namely `ProcessAccess.DUPLICATE_SAME_ACCESS` for example). However, it could lead to some confusion/errors if you use it the other way around.

So, do we want to change it now, or will we use the enum only in this particular way?

acquire/dynamic/windows/types.py
```diff
@@ -53,6 +53,27 @@ class ErrorCode(IntEnum):
...
+    key = "Key"
+    mutant = "Mutant"
+    process = "Process"
+    section = "Section"
+    semaphore = "Semaphore"
+    thread = "Thread"
+    token = "Token"
+    tp_worker_factory = "TpWorkerFactory"
+    wait_completion_packet = "WaitCompletionPacket"
+    window_station = "WindowStation"
```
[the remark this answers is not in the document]
Poeloe: Integrated it!

acquire/dynamic/windows/handles.py
```diff
@@ -154,18 +162,27 @@ def get_handles():
...
-
-        handle_name = None
-
         handle_name = get_handle_name(handle.UniqueProcessId, handle.HandleValue)
+
         if not handle_name:
             continue
 
         yield Handle(handle, handle_type, handle_name)
+    log.removeFilter(DuplicateFilter())
```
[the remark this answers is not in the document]
Poeloe: True that, changed it ;)

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,255 @@
...
+kernel32.DuplicateHandle.argtypes = (HANDLE, HANDLE, HANDLE, ctypes.POINTER(HANDLE), DWORD, BOOL, DWORD)
+
+
+class DuplicateFilter(Filter):
+    def __init__(self) -> None:
+        super().__init__()
+        self.msgs = set()
+
+    def filter(self, record: LogRecord) -> bool:
+        seen = (msg := record.getMessage()) not in self.msgs
```
pyrco, line 46: I'm not a fan of using the `:=` operator like this. It is much quicker to parse the following as a human:
```suggestion
        msg =  record.getMessage()
        seen = msg not in self.msgs
```

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,255 @@
...
+        self.msgs.add(msg)
```
pyrco, line 47: Should it always be added even if it is already present?

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,255 @@
...
+    thread = threading.Thread(target=_get_file_name_thread, args=(handle, q))
+    thread.daemon = True
+    thread.start()
+    thread.join(1.0)
+
+    result = None
+    if not thread.is_alive():
+        try:
+            result = q.get_nowait()
+        except Exception:  # noqa
```
pyrco, line 130: Make the `# noqa` more specific to what linting error should not be triggerd

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,255 @@
...
+            log.error(e)
+            return None
+        try:
+            handle = duplicate_handle(h_remote, handle)
+        except RuntimeError:
+            close_handle(h_remote)
+            return None
+
+    q = Queue()
+    thread = threading.Thread(target=_get_file_name_thread, args=(handle, q))
```
pyrco, line 121: Why is this done in a thread?

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,255 @@
...
+        handle_name = get_handle_name(handle.UniqueProcessId, handle.HandleValue)
+
+        if not handle_name:
+            continue
+
+        yield Handle(handle, handle_type, handle_name)
+    log.removeFilter(duplicate_filter)
+
+
+def open_process(pid: int) -> int:
```
pyrco, line 174: When reading through the code it is nicer to have this appear above the function that uses it. Could you move it above `get_handle_name()`?

pyrco (changes requested):

acquire/dynamic/windows/types.py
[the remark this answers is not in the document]
pyrco: I'd make it a separate enum. I found there is at least another option (even 2 in WIN2000):
```
class DuplicateHandleOptions(IntEnum):
    DUPLICATE_CLOSE_SOURCE = 0x00000001
    DUPLICATE_SAME_ACCESS = 0x00000002
    DUPLICATE_SAME_ATTRIBUTES = 0x00000003
```

The last one, `DUPLICATE_SAME_ATTRIBUTES` I could only find for WIN2000, so I'm not sure it should be included.

acquire/dynamic/windows/handles.py
Poeloe: Added a comment to explain why a thread is used here ;)

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,237 @@
...
+        else:
+            raise RuntimeError(hex(result))
+
+
+def get_handle_name(pid: int, handle: SYSTEM_HANDLE_TABLE_ENTRY_INFO_EX):
+    """Return handle name"""
+
+    def get_file_name_thread(h_file: int, q: queue.Queue):
+        iob = IO_STATUS_BLOCK()
+        file_name_information = ctypes.create_string_buffer(0x1000)
```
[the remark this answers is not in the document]
pyrco: Don't know why GitHub sets this comment as outdated, but this magic number could do with some comment about why it is `0x1000` and not another number :wink:

acquire/dynamic/windows/handles.py
```diff
@@ -0,0 +1,259 @@
...
+    # No valid handle could be obtained, display the error code
+    if h_process == 0:
+        raise OpenProcessError(f"OpenProcess Error: 0x{error:x} [pid: {pid}]")
+
+    return h_process
+
+
+def _get_file_name_thread(h_file: HANDLE, q: Queue):
+    iob = IO_STATUS_BLOCK()
+    file_name_information = ctypes.create_string_buffer(0x1000)
```
pyrco, line 122: As GitHub hides the original conversation as it is outdated somehow, but this magic number could do with some comment about why it is 0x1000 and not another number wink

pyrco (changes requested):

acquire/dynamic/windows/handles.py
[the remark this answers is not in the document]
Poeloe: Implemented it differently, now it should be more clear

acquire/dynamic/windows/exceptions.py
```diff
@@ -8,3 +8,7 @@ class NoMoreEntriesError(Exception):
 
 class HandleNotClosedSuccessfullyError(Exception):
     pass
+
+
+class OpenProcessError(Exception):
```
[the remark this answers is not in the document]
Poeloe: Like so?

acquire/dynamic/windows/types.py
```diff
@@ -32,6 +192,13 @@ def from_str(cls, value: str) -> None:
         return destination
 
 
+class PUBLIC_OBJECT_TYPE_INFORMATION(ctypes.Structure):
+    _fields_ = [
+        ("Name", UNICODE_STRING),
+        ("Reserved", ULONG * 22),
+    ]
+
```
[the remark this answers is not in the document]
Poeloe: Done

acquire/dynamic/windows/types.py
```diff
@@ -1,12 +1,172 @@
...
+    _fields_ = [
+        ("NumberOfHandles", ULONG_PTR),
+        ("Reserved", ULONG_PTR),
+        ("Handles", SYSTEM_HANDLE_TABLE_ENTRY_INFO_EX * 1),
+    ]
+
+
+class FILE_NAME_INFORMATION(ctypes.Structure):
+    _fields_ = [("FileNameLength", ULONG), ("FileName", WCHAR * 1)]
+
```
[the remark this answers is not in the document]
Poeloe: Now it is ;)

pyrco (approved):

Schamper (approved):

PR #26, opened 2022-11-23
Powershell artefacts in default/full profiles

cecinestpasunepipe:
(DIS-1533)


Schamper:
Can also add it to minimal, IMO. It's usually very small.

PR #29, opened 2022-11-25
Fix Qwinsta

cecinestpasunepipe:
(DIS-1672)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -14,9 +14,9 @@ from collections import defaultdict
 from pathlib import Path
 
 from dissect.target import Target, exceptions
-from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.filesystems import ntfs
 from dissect.target.helpers import fsutil
+from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.plugins.os.windows import iis
 from dissect.target.plugins.os.windows.log import evt, evtx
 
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -404,11 +404,15 @@ def get_spec_additions(cls, target):
...
-    SPEC = [
-        ("command", (["qwinsta", "/VM"], "win-rd-sessions")),
-    ]
     EXEC_ORDER = ExecutionOrder.BOTTOM
 
+    @classmethod
+    def get_spec_additions(cls, target):
+        qwinsta = subprocess.run(["where", "qwinsta.exe"], capture_output=True, text=True).stdout.rstrip("\n")
+        return [
+            ("command", ([qwinsta, "/VM"], "win-rdp-sessions")),
```
Schamper, line 413: Just a question, was it also a bug that the previous command did `win-rd-sessions` instead of `win-rdp-sessions`?

cecinestpasunepipe: Not intentional but I keep typing 'rdp' instead of 'rd' and even Google corrects my search automatically (just a test). So I think we better change this yeah.

Schamper: But I assume it's functionally the same?

pyrco: Yes functionally the same, but it the output of the command will now be written to `win-rdp-sessions` instead of `win-rd-sessions`. I guess that's fine, as the module is also called `WinRDPSessions` and there is not yet a plugin on the dissect.target side that consumes this file.

acquire/acquire.py
```diff
@@ -404,11 +404,15 @@ def get_spec_additions(cls, target):
...
 class WinRDPSessions(Module):
     DESC = "Windows Remote Desktop session information"
-    SPEC = [
-        ("command", (["qwinsta", "/VM"], "win-rd-sessions")),
-    ]
     EXEC_ORDER = ExecutionOrder.BOTTOM
 
+    @classmethod
+    def get_spec_additions(cls, target):
+        qwinsta = subprocess.run(["where", "qwinsta.exe"], capture_output=True, text=True).stdout.rstrip("\n")
```
pyrco, line 411: Technically it could return multiple lines if the utility is present in multiple dirs in `%path%` and current dir.

pyrco (approved):

PR #33, opened 2023-01-04
Collect proc and sys filesystems

Horofic:
(DIS-1449)
(DIS-1450)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -257,7 +257,7 @@ class Sys(Module):
     @classmethod
     def _run(cls, target: Target, collector: Collector):
         if not Path("/sys").exists() or sys.platform == "win32":
-            log.error("/sys is unavailable or acquire is not running on a Unix-like system! Skipping...")
+            log.error("/sys is unavailable! Skipping...")
             return
 
         spec = [("dir", "/sys")]
@@ -278,8 +278,8 @@ class Proc(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
-        if not Path("/proc").exists() or sys.platform == "win32":
-            log.error("/proc is unavailable or acquire is not running on a Unix-like system! Skipping...")
+        if not Path("/proc").exists():
+            log.error("/proc is unavailable! Skipping...")
             return
 
         spec = [("dir", "/proc")]
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -15,7 +15,7 @@ from collections import defaultdict
 from pathlib import Path
 
 from dissect.target import Target, exceptions
-from dissect.target.filesystems import ntfs, dir
+from dissect.target.filesystems import dir, ntfs
 from dissect.target.helpers import fsutil
 from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.plugins.os.windows import iis
@@ -251,7 +251,7 @@ class Module:
 @register_module("--sys")
 @local_module
 class Sys(Module):
-    DESC = "Procfs files (live systems only)"
+    DESC = "Sysfs files (live systems only)"
     EXEC_ORDER = ExecutionOrder.BOTTOM
 
     @classmethod
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -248,6 +248,42 @@ class Module:
         pass
 
 
+@register_module("--sys")
+@local_module
+class Sys(Module):
+    DESC = "Procfs files (live systems only)"
+    EXEC_ORDER = ExecutionOrder.BOTTOM
+
+    @classmethod
+    def _run(cls, target: Target, collector: Collector):
+        spec = [("dir", "/sys")]
+
+        sysfs = DirectoryFilesystem(Path("/sys"))
+
+        target.filesystems.add(sysfs)
+        target.fs.mount("/sys", sysfs)
+
+        collector.collect(spec, follow=False, volatile=True)
+
+
+@register_module("--proc")
+@local_module
+class Proc(Module):
+    DESC = "Procfs files (live systems only)"
+    EXEC_ORDER = ExecutionOrder.BOTTOM
+
+    @classmethod
+    def _run(cls, target: Target, collector: Collector):
+        spec = [("dir", "/proc")]
+
+        procfs = DirectoryFilesystem(Path("/proc"))
+
+        target.filesystems.add(procfs)
+        target.fs.mount("/proc", procfs)
+
+        collector.collect(spec, follow=False, volatile=True)
+
+
 @register_module("-n", "--ntfs")
 class NTFS(Module):
     DESC = "NTFS filesystem metadata"
@@ -420,6 +456,19 @@ class WinRDPSessions(Module):
             ("command", ([qwinsta, "/VM"], "win-rdp-sessions")),
         ]
 
+    @classmethod
+    def get_spec_additions(cls, target):
+        # where.exe instead of where, just in case the client runs in PS instead of CMD
+        # by default where hides qwinsta on 32-bit systems because qwinsta is only 64-bit, but with recursive /R search
+        # we can still manage to find it and by passing the exact path Windows will launch a 64-bit process
+        # on systems capable of doing that.
+        qwinsta = subprocess.run(
+            ["where.exe", "/R", os.environ["WINDIR"], "qwinsta.exe"], capture_output=True, text=True
+        ).stdout.split("\n")[0]
+        return [
+            ("command", ([qwinsta, "/VM"], "win-rdp-sessions")),
+        ]
+
 
 @register_module("--winpmem")
 @local_module
@@ -483,8 +532,8 @@ class WinMemDump(Module):
                         )
                         return
 
-            collector.output.write_entry(mem_dump_output_path, mem_dump_path)
-            collector.output.write_entry(mem_dump_errors_output_path, mem_dump_errors_path)
+            collector.output.write_entry(mem_dump_output_path, entry=mem_dump_path)
+            collector.output.write_entry(mem_dump_errors_output_path, entry=mem_dump_errors_path)
             collector.report.add_command_collected(cls.__name__, command_parts)
             mem_dump_path.unlink()
             mem_dump_errors_path.unlink()
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -256,7 +256,7 @@ class Sys(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
-        if not Path("/sys").exists() or sys.platform == "win32":
+        if not Path("/sys").exists():
             log.error("/sys is unavailable! Skipping...")
             return
 
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -256,16 +256,16 @@ class Sys(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
+        if not Path("/sys").exists() or sys.platform == "win32":
+            log.error("/sys is unavailable or acquire is not running on a Unix-like system! Skipping...")
+            return
+
         spec = [("dir", "/sys")]
 
         sysfs = dir.DirectoryFilesystem(Path("/sys"))
 
         target.filesystems.add(sysfs)
         target.fs.mount("/sys", sysfs)
-
-        if not target.fs.exists("/sys") or sys.platform == "win32":
-            log.error("/sys is unavailable or acquire is not running on a Unix-like system! Skipping...")
-            return
 
         collector.collect(spec, follow=False, volatile=True)
 
@@ -278,16 +278,15 @@ class Proc(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
-        spec = [("dir", "/proc")]
+        if not Path("/proc").exists() or sys.platform == "win32":
+            log.error("/proc is unavailable or acquire is not running on a Unix-like system! Skipping...")
+            return
 
+        spec = [("dir", "/proc")]
         procfs = dir.DirectoryFilesystem(Path("/proc"))
 
         target.filesystems.add(procfs)
         target.fs.mount("/proc", procfs)
-
-        if not target.fs.exists("/proc") or sys.platform == "win32":
-            log.error("/proc is unavailable or acquire is not running on a Unix-like system! Skipping...")
-            return
 
         collector.collect(spec, follow=False, volatile=True)
 
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -15,7 +15,7 @@ from collections import defaultdict
 from pathlib import Path
 
 from dissect.target import Target, exceptions
-from dissect.target.filesystems import ntfs
+from dissect.target.filesystems import ntfs, dir
 from dissect.target.helpers import fsutil
 from dissect.target.loaders.remote import RemoteStreamConnection
 from dissect.target.plugins.os.windows import iis
@@ -258,7 +258,7 @@ class Sys(Module):
     def _run(cls, target: Target, collector: Collector):
         spec = [("dir", "/sys")]
 
-        sysfs = DirectoryFilesystem(Path("/sys"))
+        sysfs = dir.DirectoryFilesystem(Path("/sys"))
 
         target.filesystems.add(sysfs)
         target.fs.mount("/sys", sysfs)
@@ -276,7 +276,7 @@ class Proc(Module):
     def _run(cls, target: Target, collector: Collector):
         spec = [("dir", "/proc")]
 
-        procfs = DirectoryFilesystem(Path("/proc"))
+        procfs = dir.DirectoryFilesystem(Path("/proc"))
 
         target.filesystems.add(procfs)
         target.fs.mount("/proc", procfs)
@@ -442,19 +442,6 @@ class WinArpCache(Module):
 class WinRDPSessions(Module):
     DESC = "Windows Remote Desktop session information"
     EXEC_ORDER = ExecutionOrder.BOTTOM
-
-    @classmethod
-    def get_spec_additions(cls, target):
-        # where.exe instead of where, just in case the client runs in PS instead of CMD
-        # by default where hides qwinsta on 32-bit systems because qwinsta is only 64-bit, but with recursive /R search
-        # we can still manage to find it and by passing the exact path Windows will launch a 64-bit process
-        # on systems capable of doing that.
-        qwinsta = subprocess.run(
-            ["where.exe", "/R", os.environ["WINDIR"], "qwinsta.exe"], capture_output=True, text=True
-        ).stdout.split("\n")[0]
-        return [
-            ("command", ([qwinsta, "/VM"], "win-rdp-sessions")),
-        ]
 
     @classmethod
     def get_spec_additions(cls, target):
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -263,6 +263,10 @@ class Sys(Module):
         target.filesystems.add(sysfs)
         target.fs.mount("/sys", sysfs)
 
+        if not target.fs.exists("/sys") or sys.platform == "win32":
+            log.error("/sys is unavailable or acquire is not running on a Unix-like system! Skipping...")
+            return
+
         collector.collect(spec, follow=False, volatile=True)
 
 
@@ -280,6 +284,10 @@ class Proc(Module):
 
         target.filesystems.add(procfs)
         target.fs.mount("/proc", procfs)
+
+        if not target.fs.exists("/proc") or sys.platform == "win32":
+            log.error("/proc is unavailable or acquire is not running on a Unix-like system! Skipping...")
+            return
 
         collector.collect(spec, follow=False, volatile=True)
 
@@ -1389,7 +1397,6 @@ class ActivitiesCache(Module):
 @module_arg("--ext-to-hash", action="append", help="Hash only files with the extensions provided")
 @module_arg("--glob-to-hash", action="append", help="Hash only files that match provided glob")
 class FileHashes(Module):
-
     DESC = "file hashes"
 
     DEFAULT_HASH_FUNCS = (HashFunc.MD5, HashFunc.SHA1, HashFunc.SHA256)
@@ -1449,7 +1456,6 @@ class FileHashes(Module):
             extensions = cls.DEFAULT_EXTENSIONS
 
         if cli_args.dir_to_hash or cli_args.glob_to_hash:
-
             if cli_args.glob_to_hash:
                 path_selectors.extend([("glob", glob) for glob in cli_args.glob_to_hash])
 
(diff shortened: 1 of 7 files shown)


acquire/utils.py
```diff
@@ -9,9 +10,91 @@
...
+
+    def _read(self, offset: int, length: int) -> bytes:
+        prev_pos = None
+
+        while True:
+            if self.is_writeonly:
+                break
+
+            prev_pos = self.tell()
+            self.buf += os.read(self.fd, 4096)
```
Schamper, line 64:
```suggestion
            result.append(os.read(self.fd, min(4096, length)))
```
?

acquire/utils.py
```diff
@@ -9,9 +10,91 @@
...
+            self.pos = len(self.buf)
+
+            if not self.max_filesize:
+                if prev_pos == self.tell() or self.tell() >= self.max_filesize:
+                    break
+            else:
+                if prev_pos == self.tell() or self.tell():
+                    break
+
+        return self.buf
```
Schamper, line 75:
```suggestion
        return b"".join(result)
```

acquire/utils.py
```diff
@@ -9,9 +10,91 @@
...
+        while True:
+            if self.is_writeonly:
+                break
+
+            prev_pos = self.tell()
+            self.buf += os.read(self.fd, 4096)
+
+            self.pos = len(self.buf)
+
+            if not self.max_filesize:
```
Schamper, line 68: You check `if not self.max_filesize` but then proceed to check against the `max_filesize`? I think you meant to do this the other way around?

acquire/utils.py
```diff
@@ -9,9 +10,91 @@
...
+        self.max_filesize = max_filesize
+        self.fh = open(path, string_flags)
+        self.fd = self.fh.fileno()
+
+        fcntl.fcntl(self.fd, fcntl.F_SETFL, flags)
+
+        self.pos = 0
+        self.buf = b""
+
+        super().__init__()
```
Schamper, line 51:
```suggestion
        super().__init__(max_filesize)
```
Also, if you were to do this, I believe the check you currently do in the loop is done for you by `AlignedStream`.

I believe you should be able to just do something like this:

```python
class VolatileStream(AlignedStream):
    def __init__(self, path: Path, mode: str = "rb", flags: int = os.O_RDONLY | os.O_NONBLOCK | os.O_NOATIME, size: int = 1024 * 1024 * 5):
        self.fh = path.open(mode)
        self.fd = self.fh.fileno()
        fcntl.fcntl(self.fd, fcntl.F_SETFL, flags)

        st_mode = os.fstat(self.fd).st_mode
        write_only = (st_mode & (S_IRUSR | S_IRGRP | S_IROTH)) == 0

        super().__init__(0 if write_only else size)

    def seek(self, pos: int, whence: int = io.SEEK_SET) -> int:
        raise UnsupportedOperation("VolatileStream is not seekable")

    def seekable(self) -> bool:
        return False

    def _read(self, offset: int, length: int) -> bytes:
        return os.read(self.fd, min(length, self.size - offset))
```

acquire/utils.py
```diff
@@ -60,7 +143,7 @@ def create_argument_parser(profiles: Dict, modules: Dict) -> argparse.ArgumentPa
     parser.add_argument(
         "-ot",
         "--output-type",
-        choices=OUTPUTS.keys(),
+        choices=acquire.outputs.OUTPUTS.keys(),
```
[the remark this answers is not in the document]
Horofic: Changed this due to circular imports. This is no longer the case, so this change can be reverted.

acquire/utils.py
```diff
@@ -139,6 +222,25 @@ def parse_acquire_args(parser: argparse.ArgumentParser, config_defaults: Optiona
     return command_line_args
 
 
+def get_pid() -> int:
```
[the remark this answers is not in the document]
Horofic: Reason was to keep os imports to a minimum. Though, these functions do not get used in this version of the implementation anymore. Thus, will remove these entirely

acquire/collector.py
```diff
@@ -353,6 +391,10 @@ def collect_path(
             is_file = path.is_file()
             is_symlink = path.is_symlink()
         except OSError as error:
+            if error.errno == errno.EINVAL:
+                # Some files can produce an invalid argument error when stat-ed.
+                # If this is the case. We still want to attemd to collect the file.
+                self.collect_file(path, module_name=module_name, follow=follow, volatile=volatile)
```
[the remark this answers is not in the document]
Horofic: I did some testing, and it seems this does seem to get triggered anymore. Ill remove this in the new commit

acquire/acquire.py
```diff
@@ -248,6 +248,42 @@ def _run(cls, target, collector):
...
 
+@register_module("--sys")
+@local_module
+class Sys(Module):
+    DESC = "Sysfs files (live systems only)"
+    EXEC_ORDER = ExecutionOrder.BOTTOM
+
+    @classmethod
+    def _run(cls, target: Target, collector: Collector):
+        spec = [("dir", "/sys")]
```
Schamper, line 259:
```suggestion
        if sys.platform == "win32":
            log.error("/sys collection can only run Unix-like systems! Skipping...")
            return

        spec = [("dir", "/sys")]
```
Maybe add something like this? Also to the proc one. Or maybe make it more specific rather than checking for win32, whatever you think is best :)

acquire/collector.py
```diff
@@ -331,11 +355,29 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
+        result = "OK"
+        log.info("- Collecting symlink entry %s: %s", path, result)
+
     def collect_path(
         self,
         path: Union[str, fsutil.TargetPath],
         seen_paths: Optional[Set] = None,
         module_name: Optional[str] = None,
+        follow: Optional[bool] = None,
+        volatile: Optional[bool] = None,
```
Schamper, line 380: Why not give these sane default values?

acquire/collector.py
```diff
@@ -372,21 +414,16 @@ def collect_path(
...
         elif is_file:
-            self.collect_file(path, module_name=module_name)
-        elif is_symlink:
-            self.report.add_path_failed(module_name, path)
-            log.error(
-                "- Can't collect %s (symlink to %s) in module %s",
-                path,
-                path.get().readlink(),
-                module_name,
-            )
```
Schamper, line 386: This error can still happen when there's a symlink that can't be resolved. In that case I think you'd also want to colelct the symlink itself (but still log an error/warning).

acquire/collector.py
```diff
@@ -372,21 +414,16 @@ def collect_path(
...
+            self.collect_file(path, module_name=module_name, follow=follow, volatile=volatile)
         else:
             self.report.add_path_failed(module_name, path)
             log.error("- Don't know how to collect %s in module %s", path, module_name)
+            return
```
Schamper, line 426: I don't think this adds anything?

acquire/outputs/base.py
```diff
@@ -1,20 +1,103 @@
...
+            fh = acquire.utils.VolatileStream(entry)
+            buf = fh.read()
+            size = size or len(buf)
+        except (OSError, PermissionError):
+            # Various OSErrors can occur here.
+            # If one does occur, we'd still like to have the corresponding entry.
+            buf = b""
+            size = 0
+
+        self.write_bytes(path, buf, size=size, entry=entry)
```
Schamper, line 99:
```suggestion
        self.write_bytes(path, buf, size, entry)
```

acquire/outputs/base.py
```diff
@@ -1,20 +1,103 @@
...
 
-    def close(self):
+    def close(self) -> None:
+        """Closes all handles of the file-like objects passed to the write function."""
```
Schamper, line 102: I don't think that's correct? Just put something like:
```suggestion
        """Closes the output."""
```

acquire/outputs/tar.py
```diff
@@ -1,12 +1,28 @@
 import io
 import tarfile
+from pathlib import Path
+from typing import BinaryIO, Optional
+
+from dissect.target.exceptions import FileNotFoundError as TargeFileNotFoundError
```
Schamper, line 6:
```suggestion
from dissect.target.exceptions import FileNotFoundError as DissectFileNotFoundError
```
`TargetFileNotFound` is technically correct but makes me think too much of a "target file", as in destination.

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
         info.uname = "root"
         info.gname = "root"
-        info.size = size
+        info.size = size or 0
+
+        if entry.is_symlink():
+            try:
+                info.type = tarfile.SYMTYPE
+                info.linkname = str(entry.readlink())
+            except (FileNotFoundError, TargeFileNotFoundError):
```
Schamper, line 79: Did you see this error happen anywhere? How can that be possible?

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
+        info.name = str(path)
         info.uname = "root"
         info.gname = "root"
-        info.size = size
+        info.size = size or 0
+
+        if entry.is_symlink():
+            try:
+                info.type = tarfile.SYMTYPE
+                info.linkname = str(entry.readlink())
```
Schamper, line 78: `readlink()` already returns a string I believe?

acquire/outputs/tar.py
Schamper: Also it's weird that you can get a real `FileNotFoundError` here, that should be caught at the `dissect.target` level. Maybe look into that?

Schamper (changes requested):

acquire/collector.py
Horofic: Missed these ones! Thanks.

acquire/outputs/tar.py
Horofic: Seems like this has changed during `dissect.target` versions. After some testing i'm only encountering `dissect.target`'s `FileNotFoundError`.

acquire/outputs/tar.py
Horofic: Changed this to just `from dissect.target.exceptions import FileNotFoundError`

acquire/collector.py
Horofic: I believe this code-path triggered when the entry is a symlink. Which causes no symlinks to be collected. I will add some error checking in the `collect_symlink()`

acquire/acquire.py
Horofic: Good one! Made the check somewhat more specific. I'd like to acquire `/proc` and `/sys` from WSL child targets in the future. Though, that should be a separate PR / ticket

acquire/collector.py
```diff
@@ -331,13 +350,39 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
+            self.report.add_symlink_collected(module_name, path)
+
+            result = "OK"
+            log.info("- Collecting symlink entry %s: %s", path, result)
+        except Exception:
+            self.report.add_symlink_failed(module_name, path)
+            log.error(
+                "- Failed to collect symlink %s (symlink to %s) in module %s",
+                path,
+                path.get().readlink(),
```
Schamper, line 373:
```suggestion
                path.readlink(),
```

acquire/collector.py
```diff
@@ -372,18 +417,20 @@ def collect_path(
...
-        elif is_symlink:
-            self.report.add_path_failed(module_name, path)
-            log.error(
-                "- Can't collect %s (symlink to %s) in module %s",
-                path,
-                path.get().readlink(),
-                module_name,
-            )
+            self.collect_file(path, module_name=module_name, follow=follow, volatile=volatile)
+        # elif is_symlink:
```
Schamper, line 426: Remove commented code.

acquire/collector.py
```diff
@@ -331,13 +350,39 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
+            )
+
     def collect_path(
         self,
-        path: Union[str, fsutil.TargetPath],
+        path: Union[str, Path],
         seen_paths: Optional[Set] = None,
         module_name: Optional[str] = None,
+        follow: Optional[bool] = True,
+        volatile: Optional[bool] = False,
```
Schamper, line 384:
```suggestion
        follow: bool = True,
        volatile: bool = False,
```

acquire/collector.py
```diff
@@ -331,13 +350,39 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
             else:
                 self.report.add_glob_collected(module_name, pattern)
 
+    def collect_symlink(
+        self,
+        path: Path,
+        module_name: Optional[str] = None,
+        follow: Optional[bool] = True,
+        volatile: Optional[bool] = False,
```
Schamper, line 358:
```suggestion
        follow: bool = True,
        volatile: bool = False,
```

acquire/collector.py
```diff
@@ -331,13 +350,39 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
+    ) -> None:
+        try:
+            outpath = self.create_outpath(str(path))
+
+            self.output.write_bytes(outpath, b"", entry=path, size=0)
```
Schamper, line 363: The typehint for `entry` is `FilesystemEntry` but you pass in the `TargetPath` here.

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
             size = fh.tell()
             fh.seek(offset)
 
         info = self.tar.tarinfo()
-        info.name = path
+        info.name = str(path)
         info.uname = "root"
         info.gname = "root"
-        info.size = size
+        info.size = size or 0
```
Schamper, line 73: I believe this should fix the allow zero thing in `dissect.target`.

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
+            except FileNotFoundError:
+                pass
+
+        if stat:
+            info.mtime = stat.st_mtime
 
         self.tar.addfile(info, fh)
 
     def close(self):
+        """Closes the opened tar file and opened file-like objects."""
```
Schamper, line 99:
```suggestion
        """Closes the tar file."""
```

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
         info.uname = "root"
         info.gname = "root"
-        info.size = size
+        info.size = size or 0
+
+        if entry.is_symlink():
+            try:
+                info.type = tarfile.SYMTYPE
+                info.linkname = entry.readlink().as_posix()
+            except FileNotFoundError:
```
Schamper, line 79: I don't think this can error.

acquire/outputs/tar.py
```diff
@@ -26,23 +42,61 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
+                pass
+
+            try:
+                stat = entry.lstat()
+            except FileNotFoundError:
```
Schamper, line 84: Same here.

Schamper (changes requested):

acquire/outputs/tar.py
Horofic: This indeed does not seem error anymore.

acquire/outputs/tar.py
Horofic: This still seems to occur when something other that `--proc` or `--sys` is provided. So using `--proc --sys -p minimal` will cause some entries to be hidden without `ignore_zeros`. But `--proc --sys` works just fine.

acquire/acquire.py
```diff
@@ -248,6 +248,49 @@ def _run(cls, target, collector):
...
 
+@register_module("--sys")
+@local_module
+class Sys(Module):
+    DESC = "Sysfs files (live systems only)"
+    EXEC_ORDER = ExecutionOrder.BOTTOM
+
+    @classmethod
+    def _run(cls, target: Target, collector: Collector):
+        if not Path("/sys").exists() or sys.platform == "win32":
```
Schamper, line 259:
```suggestion
        if not Path("/sys").exists():
```
Just guessing here that the check for `win32` doesn't work, because `/sys` will never exist on Windows so that check will never be done.

acquire/acquire.py
```diff
@@ -248,6 +248,49 @@ def _run(cls, target, collector):
...
+
+@register_module("--proc")
+@local_module
+class Proc(Module):
+    DESC = "Procfs files (live systems only)"
+    EXEC_ORDER = ExecutionOrder.BOTTOM
+
+    @classmethod
+    def _run(cls, target: Target, collector: Collector):
+        if not Path("/proc").exists() or sys.platform == "win32":
```
Schamper, line 281:
```suggestion
        if not Path("/proc").exists():
```
Same here.

acquire/collector.py
```diff
@@ -331,13 +350,40 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
+    def collect_symlink(
+        self,
+        path: Path,
+        module_name: Optional[str] = None,
+        follow: bool = True,
+        volatile: bool = False,
+    ) -> None:
+        try:
+            outpath = self.create_outpath(str(path))
+            entry = path.get()
```
Schamper, line 362: Now this code path is the only one actually passing in a `FilesystemEntry`, all other code paths (including the memory dump one) are passing in a `Path`-like object (either a real `Path` or a `TargetPath`).

acquire/outputs/tar.py
```diff
@@ -1,12 +1,28 @@
...
 
 
 class TarOutput(Output):
-    def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwargs):
+    """Tar archive acquire output format. Output can be compressed and/or encrypted.
+
+    Args:
+        path: The path to write the tar archive to.
+        compress: Whether to compress the tar archive.
+        encrypt:  Whether to encrypt the tar archive.
```
Schamper, line 19:
```suggestion
        encrypt: Whether to encrypt the tar archive.
```

acquire/outputs/tar.py
```diff
@@ -26,23 +42,51 @@ def __init__(self, path, compress=False, encrypt=False, public_key=None, **kwarg
...
+        size: Optional[int] = None,
+        entry: Optional[FilesystemEntry] = None,
+    ) -> None:
+        """Write a filesystem entry or file-like object to a tar file.
+
+        Args:
+            path: The path of the entry to write.
+            fh: The file-like object of the entry to write.
+            size: The optional file size in bytes of the entry to write.
+            entry: the optional filesystem entry of the entry to write.
```
Schamper, line 58:
```suggestion
            entry: The optional filesystem entry of the entry to write.
```

Schamper: Also noting that this might still be a wrong type hint, see other comment.

Schamper (changes requested):

acquire/outputs/tar.py
Horofic: I believe the `path` variable here primarily gets used to constructs (arbitrary) paths to map entires into their respective output format, like `$rootfs$/etc/hostname` to `/etc/hostname`. I'am thinking about changing `path` to `output_path` and changing the docstring as well to better reflect this.

I will fix the `Path` and `FilesystemEntry` mismatch  in the process as well. Of the other functions as well.

acquire/acquire.py
Schamper: The recent change doesn't address this.

acquire/collector.py
```diff
@@ -331,13 +351,40 @@ def collect_glob(self, pattern: str, module_name: Optional[str] = None) -> None:
...
 
+    def collect_symlink(
+        self,
+        path: Path,
+        module_name: Optional[str] = None,
+        follow: bool = True,
+        volatile: bool = False,
+    ) -> None:
+        try:
+            outpath = self.create_outpath(str(path))
```
Schamper, line 362:
```suggestion
            outpath = self.create_outpath(path)
```

acquire/utils.py
```diff
@@ -1,19 +1,64 @@
...
+    """
+
+    def __init__(
+        self,
+        path: Path,
+        mode: str = "rb",
+        flags: int = os.O_RDONLY | os.O_NONBLOCK,
+        size: int = 1024 * 1024 * 5,
+    ):
+        if sys.platform != "darwin":
```
Schamper, line 39:
```suggestion
        if hasattr(os, "O_NOATIME"):
```

Schamper (changes requested):

acquire/collector.py
```diff
@@ -362,9 +366,17 @@ def collect_symlink(
             outpath = self.create_outpath(path)
             entry = path.get()
 
-            self.output.write_bytes(outpath, b"", 0, entry)
-            self.report.add_symlink_collected(module_name, path)
+            if follow:
+                resolved_path = path.resolve()
+                if resolved_path.is_file():
+                    self.collect_file(resolved_path, size, module_name=module_name, volatile=volatile)
```
Schamper, line 372: You didn't end up doing this logic in `collect_path`? It's kinda weird now, because it generally doesn't make sense to pass a size to collect a symlink, because it doesn't have a size.

Horofic: I find that by moving the logic to the `collect_symlink` it made `collect_path` cleaner. You are right about the `size`. After some testing the size didn't end up being needed. So the parameter can be removed. This should make  `collect_symlink` make more sense.

acquire/utils.py
```diff
@@ -34,16 +41,15 @@ def __init__(
         self,
         path: Path,
         mode: str = "rb",
-        flags: int = os.O_RDONLY | os.O_NONBLOCK,
+        # Windows and Darwin systems dont have O_NOATIME or O_NONBLOCK. We still want to add them available.
```
Schamper, line 44:
```suggestion
        # Windows and Darwin systems don't have O_NOATIME or O_NONBLOCK. Add them if they are available.
```

Schamper (changes requested):

Schamper (approved):

acquire/utils.py
```diff
@@ -8,12 +8,63 @@
...
+        mode: Mode string to open the file-handle with. Such as "rt" and "rb".
+        flags: Flags to open the file-descriptor with.
+        size: The maximum size of the stream. None if unknown.
+    """
+
+    def __init__(
+        self,
+        path: Path,
+        mode: str = "rb",
+        # Windows and Darwin systems don't have O_NOATIME or O_NONBLOCK. Windows and Darwin systems don't have O_NOATIME or O_NONBLOCK. Add them if they are available.
```
Schamper, line 44:
```suggestion
        # Windows and Darwin systems don't have O_NOATIME or O_NONBLOCK. Add them if they are available.
```

Schamper (changes requested):

PR #37, opened 2023-02-03
Improve upload order to put log files last

Miauwkeru:
(DIS-1803)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -13,7 +13,7 @@ import urllib.parse
 import urllib.request
 from collections import defaultdict
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
 from dissect.target import Target, exceptions
 from dissect.target.filesystems import ntfs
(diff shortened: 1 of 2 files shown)


acquire/acquire.py
```diff
@@ -1948,20 +1949,25 @@ def acquire_children_and_targets(target: Target, args: argparse.Namespace):
...
-                child_files = acquire_target(child_target, args)
+                child_files, child_log_file = acquire_target(child_target, args)
                 files.extend(child_files)
+                log_files.extend(child_log_file)
             except Exception:
                 log.exception("Failed to acquire child target")
                 continue
 
+    # The main log file is always first, so we reverse the list to put that one at the end.
+    log_files = [log_file for log_file in log_files if log_file]
```
pyrco, line 1960: If I see it correct, if `args.log_path` is provided, everything is logged to the same `log_file`, so this list will have multiple instances of the same `log_file`. Maybe good to de-dup the list.

Another thing I'm not sure of is what happens if the file does not exist, as previously when `acquire_target` failed at the start, the list of files it returned was empty, now it returns the `log_file` `Path`.

pyrco (changes requested):

pyrco (approved):

PR #40, opened 2023-02-20
Update path to IIS module

Schamper:
(DIS-1840)

Necessary change since https://github.com/fox-it/dissect.target/pull/127


Schamper:
> I'll ack it if it gets through the pipeline

I believe that requires a new stable of `dissect.target`. This should be tightly coordinated with the next release, because this PR wil need to be merged before making a new release of `acquire`. Otherwise stuff will break.

pyrco:
Or we can change the dependency to 
```
dissect.target[full]>=3.0.dev,<4.0.dev
```
As acquire is now so tightly coupled and released with the dissect  projects I think this is ok.

PR #44, opened 2023-03-10
Acquire McAfee SQLite-based logs as well

cecinestpasunepipe:
Adds the SQLite3-based logs to AV as well.

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -830,7 +830,7 @@ class AV(Module):
         ("dir", "sysvol/ProgramData/McAfee/Endpoint Security/Logs_Old"),
         ("dir", "sysvol/ProgramData/Mcafee/VirusScan"),
         ("dir", "sysvol/ProgramData/McAfee/Endpoint Security/Logs"),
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -830,6 +830,7 @@ class AV(Module):
         ("dir", "sysvol/ProgramData/McAfee/Endpoint Security/Logs_Old"),
         ("dir", "sysvol/ProgramData/Mcafee/VirusScan"),
         ("dir", "sysvol/ProgramData/McAfee/Endpoint Security/Logs"),
+        ("glob", "sysvol/ProgramData/McAfee/MSC/Logs/*.log"),
```
Schamper, line 833: Any particular reason why we wouldn't collect the entire `Logs` directory? Like the other lines do.

cecinestpasunepipe: Not sure what else there is in that directory, might be redundant, but okay.

Schamper: Better to collect too much than not enough :smile:

Schamper (approved):

PR #46, opened 2023-03-16
Derive report name from output name if available

### LINKED ISSUE #45: Feature Request: allow custom report names
Currently when `acquire` is ran, the tool writes a report to the output directory prefixed by the hostname of the machine, a timestamp and a suffix of `.report.json`. However, whenever a custom output file is specified, the report does not adhere to this same naming scheme. I was looking for the option, and I could only find an option for disabling the report all together.

This mainly becomes tedious when every single machine that is `acuire`'d has the exact same hostname, making reports hard to distinguish from each other.

Would it be possible to either have the report be written to a similarly named file, or have an extra option which lets users specify their own report path?

Schamper:
Fixes #45 

Specifying `-o /tmp` will use the same old behaviour, e.g. result in `/tmp/DESKTOP-46R9IUD_20230316170910.report.json`. Specifying `-o /tmp/test.tar` will instead result in `/tmp/test.tar.report.json`. Dealing with the possible extra extension will be hard because you don't necessarily know if it should be part of the file name (e.g. extension left out, but filename contains an AD domain). Checking for common extensions seems a bit overkill for this.

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1718,13 +1718,14 @@ def acquire_target(target: Target, args: argparse.Namespace, output_ts: Optional
             **collection_report_serialized,
         }
 
-        report_output_dir = output_path if output_path.is_dir() else output_path.parent
-        report_file_path = persist_execution_report(
-            output_dir=report_output_dir,
-            prefix=target.name,
-            timestamp=output_ts,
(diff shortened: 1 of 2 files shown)


Schamper:
Pipeline will succeed once #40 is merged.

PR #49, opened 2023-03-24
Print an "EXPERIMENTAL" warning when handling an unsupported system

Miauwkeru:
(DIS-1612)﻿

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -313,11 +313,10 @@ class NTFS(Module):
 
     @classmethod
     def collect_usnjrnl(cls, collector: Collector, fs, name: str) -> None:
-        usnjrnl_path = fs.path("$Extend/$Usnjrnl")
-
         try:
+            usnjrnl_path = fs.path("$Extend/$Usnjrnl:$J")
             entry = usnjrnl_path.get()
-            journal = entry.entry.open("$J")
+            journal = entry.open()
 
             i = 0
             while journal.runlist[i][0] is None:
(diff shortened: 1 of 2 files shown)


acquire/acquire.py
```diff
@@ -1551,6 +1551,18 @@ def print_volumes_overview(target):
     log.info("")
 
 
+def print_acquire_warning(target: Target) -> None:
+    if target.os != "windows":
+        log.warning("========================================== WARNING ==========================================")
+        log.warning("")
+        log.warning(
+            "The operating system '%s' is not officially supported by acquire, and should be considered experimental.",
```
Schamper, line 1559: I'd change "officially" to something a bit less fearful (because there are modules for collecting stuff from other OS's, so you could argue there is "official support"). Perhaps just saying `Operating system support for '%s' is experimental. Some artifacts may not yet be included and some features may not work as expected.`.

Maybe also add something along the lines of "Please notify upstream for any missing artifacts or features".

Schamper (changes requested):

pyrco (dismissed):

cecinestpasunepipe (dismissed):

acquire/acquire.py
```diff
@@ -1551,6 +1551,19 @@ def print_volumes_overview(target):
     log.info("")
 
 
+def print_acquire_warning(target: Target) -> None:
+    if target.os != "windows":
+        log.warning("========================================== WARNING ==========================================")
+        log.warning("")
+        log.warning(
+            "The operating system '%s' is experimental. Some artifacts may not yet be included and some features",
```
Schamper, line 1559: The operating system is not experimental, support for it is.

Schamper (approved):

PR #53, opened 2023-04-06
Skip mount points when collecting a directory

pyrco:
When collecting a directory or glob, any subdirectories within that
directory or within directories expanded from the glob, that are also
mount points, will be skipped.

(DIS-1785)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -272,7 +272,7 @@ class Sys(Module):
         target.filesystems.add(sysfs)
         target.fs.mount("/sys", sysfs)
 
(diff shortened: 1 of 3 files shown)


Schamper:
Can you explain a bit on the behaviour that this change will now have?

pyrco:
After some discussion with @Horofic this is probably not the way to go. Collecting `/sys` and `/proc` depend on collecting directories over fs boundaries. I'm working on a solution that is more specific and will just prevent following the symlinks in the `/proc/self/fd` directory, where `self` will be resolved to the actual pid.

PR #56, opened 2023-04-11
Skip collecting /proc/self/fd/* symlinks on live systems

pyrco:
One of the fds in /proc/self/fd is a symlink to the tar file in which all artefacts are collected. Collecting the file this symlink links to will result in a "self" collect, which is not desired.

On certain systems /proc/self/fd is bind mounted on some other part of the system (e.g. somewhere below /var), which will result in such a situation. Note that collecting /proc with --proc does not suffer this issue as it will not collect the file a symlink points to by design.

When a symlink is collected and its parent resolves to the same directory /proc/self/fd resolves to (the self part is also a symlink pointing to the pid of the acquire process), only the symlink will be collected and not the location it points to.

(DIS-1785)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -261,17 +261,15 @@ class Sys(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
-        if not Path("/sys").exists():
+        if not str(target.path) == "local":
+            log.error("Not running on a live system, skipping collection of /sys")
+            return
+
+        if not target.fs.path("/sys").exists():
             log.error("/sys is unavailable! Skipping...")
             return
 
         spec = [("dir", "/sys")]
-
-        sysfs = dir.DirectoryFilesystem(Path("/sys"))
-
-        target.filesystems.add(sysfs)
-        target.fs.mount("/sys", sysfs)
-
         collector.collect(spec, follow=False, volatile=True)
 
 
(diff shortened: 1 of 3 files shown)


acquire/acquire.py
```diff
@@ -1961,6 +1958,19 @@ def main():
         log.exception("Failed to load target")
         raise
 
+    if target_path == "local":
+        proc_path = Path("/proc")
+        if proc_path.exists():
+            procfs = dir.DirectoryFileproctem(proc_path)
+            target.fileproctems.add(procfs)
```
Horofic, line 1965:
```suggestion
            target.filesystems.add(procfs)
```

Horofic (changes requested):

acquire/acquire.py
```diff
@@ -261,17 +261,15 @@ class Sys(Module):
 
     @classmethod
     def _run(cls, target: Target, collector: Collector):
-        if not Path("/sys").exists():
+        if not str(target.path) == "local":
```
Schamper, line 264: Just a note, this method (and the way `/proc` and `/sys` are currently being mounted) won't work in the future with a remote agent or targetd target. Perhaps mounting of `/proc` and `/sys` should be moved to the `LocalLoader` instead.

pyrco: I made a ticket for this.

PR #66, opened 2023-06-08
add limited support for zip output (no encryption yet)

janstarke:


acquire/outputs/zip.py
```diff
@@ -0,0 +1,100 @@
+import io
+import pyzipper
```
Schamper, line 2: Any specific reason for not using `zipfile` (https://docs.python.org/3/library/zipfile.html)?

acquire/outputs/zip.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self._fh = None
+        self.path = path.with_suffix(path.suffix + ext)
+
+        if compress:
+            self.compression = pyzipper.ZIP_LZMA
+        else:
+            self.compression = pyzipper.ZIP_STORED
+
+        if encrypt:
```
Schamper, line 40: It's preferred to use `EncryptedStream` instead of the encryption of ZIP itself, so that all crypto operations are universal on all acquire output files.

Example:
https://github.com/fox-it/acquire/blob/main/acquire/outputs/tar.py#L43

acquire/outputs/zip.py
cecinestpasunepipe: Agreed, better use the standard library to avoid unnecessary dependencies. Since this is quite a fundamental issue with this PR I say further reviewing at this point is premature (because changing the library might have a big impact on the overall code).

cecinestpasunepipe (changes requested):

janstarke:
Dear all, I replaced pyzipper by zipfile

acquire/outputs/zip.py
```diff
@@ -0,0 +1,112 @@
...
+        info = zipfile.ZipInfo()
+        info.filename = output_path
+
+        # some BinaryIO objects have no size, but `zipfile` uses len() in several places,
+        # so we read the whole data first
+        try:
+            info.file_size = len(fh)
+        except:
+            fh = fh.read()
+            info.file_size = len(fh)
```
Schamper, line 90:
```suggestion
        info.file_size = size or 0
```
We determined the size just earlier.

acquire/outputs/zip.py
```diff
@@ -0,0 +1,112 @@
...
+            lstat = entry.lstat()
+            if lstat:
+                dt = datetime.fromtimestamp(lstat.st_mtime)
+                info.date_time = (dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second)
+
+        self.archive.writestr(info, fh, compress_type=self.compression)
+
+    def close(self) -> None:
+        """Closes the archive file."""
+        self.archive.close()
```
Schamper, line 112:
```suggestion
        self.archive.close()
        if self._fh:
            self._fh.close()
```

Schamper (changes requested):
Could you also run `black` for the linting? Or `tox -e fix` for running it with `tox`.

acquire/outputs/zip.py
```diff
@@ -0,0 +1,116 @@
...
+                # Windows does not have symlinks, so this must be a unixoid system
+                info.create_system = 3
+
+                """
+                The Python zipfile module accepts the 16-bit "Mode" field
+                (that stores st_mode field from struct stat, containing
+                user/group/other permissions, setuid/setgid and symlink
+                info, etc) of the ASi extra block for Unix as bits 16-31 of
+                the external_attr
+                """
```
Schamper, line 90:
```suggestion
                # The Python zipfile module accepts the 16-bit "Mode" field (that stores st_mode field from 
                # struct stat, containing user/group/other permissions, setuid/setgid and symlink info, etc) of the 
                # ASi extra block for Unix as bits 16-31 of the external_attr
```

Schamper (changes requested):

Horofic:
Hey @janstarke! First of all, thank for this contribution. It looks like the current implementation does not play nice with the acquisition of files in `/proc` and `/sys`, which was added some time ago in https://github.com/fox-it/acquire/pull/94 and https://github.com/fox-it/acquire/pull/33. I will take a look at this the upcoming days :).

acquire/outputs/zip.py
```diff
@@ -0,0 +1,112 @@
...
+                    | stat.S_IXOTH
+                )
+                info.external_attr = unix_st_mode << 16
+
+            lstat = entry.lstat()
+            if lstat:
+                dt = datetime.fromtimestamp(lstat.st_mtime)
+                info.date_time = (dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second)
+
+        self.archive.writestr(info, fh, compress_type=self.compression)
```
Horofic, line 106:
```suggestion
        with self.archive.open(info, "w") as zfh:
            shutil.copyfileobj(fh, zfh)
```
Writing the zip-file to disk like this, allows it to play nice with file-handles that are not a `SizedBuffer`, like some of the [`VolatileStream`](https://github.com/fox-it/acquire/blob/main/acquire/volatilestream.py) file-handles

acquire/outputs/zip.py
```diff
@@ -0,0 +1,112 @@
+import io
```
Horofic, line 1:
```suggestion
import io
import shutil
```
This import is needed for suggestion below.

Horofic (changes requested):

acquire/outputs/zip.py
```diff
@@ -0,0 +1,114 @@
...
+        info.file_size = size or 0
+
+        if entry:
+            if entry.is_symlink():
+                # System which created ZIP archive, 3 = Unix; 0 = Windows
+                # Windows does not have symlinks, so this must be a unixoid system
+                info.create_system = 3
+
+                # The Python zipfile module accepts the 16-bit "Mode" field (that stores st_mode field from 
+                # struct stat, containing user/group/other permissions, setuid/setgid and symlink info, etc) of the
```
Horofic, line 86:
```suggestion
                # The Python zipfile module accepts the 16-bit "Mode" field (that stores st_mode field from
                # struct stat, containing user/group/other permissions, setuid/setgid and symlink info, etc) of the
```

Horofic (dismissed):

acquire/outputs/zip.py
```diff
@@ -0,0 +1,114 @@
...
+            self._fh = EncryptedStream(self.path.open("wb"), public_key)
+            self.archive = zipfile.ZipFile(fileobj=self._fh, mode="w", compression=self.compression, allowZip64=True)
+        else:
+            self.archive = zipfile.ZipFile(self.path, mode="w", compression=self.compression, allowZip64=True)
+
+    def write(
+        self,
+        output_path: str,
+        fh: BinaryIO,
+        entry: Optional[Union[FilesystemEntry, Path]],
```
Schamper, line 55:
```suggestion
        entry: Optional[Union[FilesystemEntry, Path]] = None,
```

cecinestpasunepipe (approved):

PR #68, opened 2023-06-15
Search for encrypted files recursively

Miauwkeru:
(DIS-1771)

### DIFF
--- a/acquire/tools/decrypter.py
+++ b/acquire/tools/decrypter.py
@@ -2,7 +2,6 @@ import argparse
 import base64
 import contextlib
 import io
-import itertools
 import json
 import logging
 import multiprocessing


acquire/tools/decrypter.py
```diff
@@ -343,19 +343,16 @@ def main():
     if not progress:
         log.info("`rich` is not installed, progress will not be shown")
 
-    if args.output and len(args.files) > 1:
+    if args.output and len(args.files) > 1 and not args.files[0].is_file():
         parser.exit("--output is only allowed when decrypting a single file")
```
pyrco, line 347: In the new situation, this is no longer correct right? I think the test should be that `args.output` is a directory if `len(args.files) > 1` or if `len(args.files) == 1 and args.files[0].is_dir()`

acquire/tools/decrypter.py
```diff
@@ -424,6 +421,18 @@ def main():
                     break
 
 
+def find_enc_files(files: list[Path]):
+    output_files = []
```
pyrco, line 425: These are not output files, it searches for encrypted files. Maybe they are used later on to construct output files, but that is not in the context of this function

pyrco (changes requested):

acquire/tools/decrypter.py
```diff
@@ -424,6 +422,18 @@ def main():
                     break
 
 
+def find_enc_files(files: list[Path]):
```
pyrco, line 425: The return type is missing.

acquire/tools/decrypter.py
```diff
@@ -343,19 +343,17 @@ def main():
...
-    if args.output and len(args.files) > 1:
-        parser.exit("--output is only allowed when decrypting a single file")
+    if args.output and args.output.is_file() and len(args.files) > 1:
+        parser.exit("--output should be a directory when decrypting multiple files.")
 
-    if not args.output:
-        for path in args.files:
-            if path.suffix != ".enc":
-                parser.exit(f"File doesn't have .enc extension: {path}")
+    files: list[Path] = find_enc_files(args.files)
```
pyrco, line 349:
```suggestion
    files = find_enc_files(args.files)
```

acquire/tools/decrypter.py
```diff
@@ -343,19 +343,17 @@ def main():
...
 
     if args.output:
-        outputs = [args.output.resolve()]
+        resolv_path = args.output.resolve()
+        outputs = [resolv_path / path.stem for path in files] if args.output.is_dir() else [resolv_path]
```
pyrco, line 353: `.stem` could lead to collisions if there are multiple input files with the same name but in different directories. But apart from recreating the whole sub-dir hierarchy of `path`, I don't see any other elegant options. Would it be worth to warn (or even error) if `len(set(ouputs)) != len(outputs)` or something like that?

pyrco (changes requested):

pyrco (approved):

PR #78, opened 2023-07-18
Add option to collect private ssh keys

Miauwkeru:
- Iterate over `/Users/*` in from_user_home for osx
- Create file_filter inside the collector
- Use a context manager to bind the module
- Add python test modules and stuff
(DIS-1869)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1358,6 +1358,11 @@ class Boot(Module):
     ]
 
 
+def private_key_filter(path: fsutil.TargetPath) -> bool:
+    with path.open("rt") as file:
+        return "PRIVATE KEY" in file.readline()
+
+
 @register_module("--home")
 class Home(Module):
     SPEC = [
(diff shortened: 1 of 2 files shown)


Zawadidone:
> @Zawadidone I did change the ssh code a lil bit more here. ':) To allow for a filter option

Nice work! FYI I am currently working on this https://github.com/fox-it/dissect.target/issues/300.

pyrco (commented):
Could you split this change set in 3 commits (and not squash them on merge):
- the new/stub `from_user_home()` changes
- the `collector.bind_module()` changes
-  the ssh private keys & file filter changes

As they are changes in different sets of fucntionality.

acquire/acquire.py
```diff
@@ -122,7 +122,13 @@ def misc_windows_user_homes(target):
...
 
 
-def from_user_home(target, path):
+def from_user_home(target: Target, path: str):
+    if target.os == "osx":
+        # Until getting the users from osx is implemented in dissect.target
+        for homedir in target.fs.path("/Users/").glob("*"):
+            yield str(homedir.joinpath(path))
+        # Return after yielding all osx users in the user directory.
+        return
```
pyrco, line 131: Would it be an idea to split this off into a `misc_osx_user_homes()`, similar as for windows? I think we also want to iterate over `/Users/*` (and `/home/*` and `/root` for other Unixes), same as we do for windows, irrespective of what comes back from `user_details()`

pyrco (changes requested):

pyrco (approved):

PR #79, opened 2023-07-24
Remove directories that are not related to tasks

Zawadidone:
These paths are already part of miscellaneous Windows artefacts https://github.com/fox-it/acquire/blob/1c1cee07900ea5985944aa2d333661dd25ec6f9b/acquire/acquire.py#L852-L853


pyrco:
Yes they are, but they are also needed when you need tasks related data. And since you can opt to not collect the Misc module paths, you could miss them. When collecting both, they should be de-duplicated in the output though, so you won't have doubles.

Zawadidone:
Okay nice feature https://github.com/fox-it/acquire/blob/8d15a39a66913f15eb64c9361296b3bdab2aac1e/acquire/collector.py#L299

PR #91, opened 2023-08-22
Add paths related to Windows memory

Zawadidone:


acquire/acquire.py
```diff
@@ -557,6 +558,36 @@ def _run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector
...
+        ("file", "sysvol/swapfile.sys"),
+        ("file", "sysvol/windows/memory.dmp"),
+        ("dir", "sysvol/windows/minidump"),
+    ]
+
+    @classmethod
+    def get_spec_additions(cls, target: Target) -> Iterator[tuple]:
+        spec = set()
+
+        resolver = ResolverPlugin(target)
```
Schamper, line 576: I believe you should just be able to do `target.resolve(path)` without having to instantiate the plugin yourself.

acquire/acquire.py
```diff
@@ -557,6 +557,34 @@ def _run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector
...
+    ]
+
+    @classmethod
+    def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Iterator[tuple]:
+        spec = set()
+
+        page_key = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Memory Management"
+        for reg_key in target.registry.iterkeys(page_key):
+            for page_path in reg_key.value("ExistingPageFiles").value:
+                spec.add(("file", target.resolve(page_path)))
```
Poeloe, line 578: Since Dissect 3.8 it is now possible to use globbing for the registry, so try changing the code to the following in order to eliminate the first for-loop:

```suggestion
        page_key = "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Memory Management\\*"
        for page_path in page_key.value("ExistingPageFiles").value:
             spec.add(("file", target.resolve(page_path)))
```

Poeloe (changes requested):
Thanks for your contribution!

Would you be able to add some general test for this new module as well?

Zawadidone:
> Thanks for your contribution!
> 
> Would you be able to add some general test for this new module as well?

What kind of test? Not one of the other modules have a test.

Poeloe:
> > Thanks for your contribution!
> > Would you be able to add some general test for this new module as well?
> 
> What kind of test? Not one of the other modules have a test.

You're right, I was confusing it with dissect.target plugins ;). So, LGTM!

Poeloe (approved):

Schamper (approved):

PR #96, opened 2023-08-29
Add MacOS applications path

Zawadidone:
References 
* https://github.com/fox-it/dissect.target/blob/b6ea46e2ed95321440df6c8c6b65803af06da859/dissect/target/plugins/os/unix/bsd/osx/_os.py#L21
* https://github.com/ForensicArtifacts/artifacts/blob/ae1baa1675e43d199721dd2e97678ac56330b5a1/data/macos.yaml#L97-L106

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1544,7 +1544,7 @@ class OSXApplicationsData(Module):
             ["/Library/Application Support"],
             from_user_home(target, "Library/Application Support"),
         ):
-            for path in target.fs.path(app_support_path).glob("*/*"):
+            for path in target.fs.path(app_support_path).glob("*/*/*"):
                 if not path.is_file():
                     continue
 
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -255,13 +255,11 @@ class Module:
     EXEC_ORDER = ExecutionOrder.DEFAULT
 
     @classmethod
-    def run(cls, target, cli_args, collector):
+    def run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector):
         desc = cls.DESC or cls.__name__.lower()
         log.info("*** Acquiring %s", desc)
 
-        collector.bind(cls)
-
-        try:
+        with collector.bind_module(cls):
             collector.collect(cls.SPEC)
 
             spec_ext = cls.get_spec_additions(target)
@@ -269,8 +267,6 @@ class Module:
                 collector.collect(list(spec_ext))
 
             cls._run(target, collector)
-        finally:
-            collector.unbind()
 
     @classmethod
     def get_spec_additions(cls, target):
@@ -1584,8 +1580,7 @@ class FileHashes(Module):
 
         specs = cls.get_specs(cli_args)
 
-        collector.bind(cls)
-        try:
+        with collector.bind_module(cls):
             start = time.time()
 
             path_hashes = collect_hashes(target, specs, path_filters=cls.DEFAULT_FILE_FILTERS)
@@ -1596,8 +1591,6 @@ class FileHashes(Module):
                 csv_compressed_bytes,
             )
             log.info("Hashing is done, %s files processed in %.2f secs", rows_count, (time.time() - start))
-        finally:
-            collector.unbind()
 
     @classmethod
     def get_specs(cls, cli_args):
@@ -1652,8 +1645,7 @@ class OpenHandles(Module):
 
         handle_types = cli_args.handle_types
 
-        collector.bind(cls)
-        try:
+        with collector.bind_module(cls):
             handles = collect_open_handles(handle_types)
             csv_compressed_handles = serialize_handles_into_csv(handles)
 
@@ -1662,8 +1654,6 @@ class OpenHandles(Module):
                 csv_compressed_handles,
             )
             log.info("Collecting open handles is done.")
-        finally:
-            collector.unbind()
 
 
 def print_disks_overview(target):
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1512,6 +1512,7 @@ class OSXApplicationsInfo(Module):
         ("glob", "Applications/*/Contents/Info.plist", from_user_home),
     ]
 
+
 @register_module("--bootbanks")
 class Bootbanks(Module):
     DESC = "ESXi bootbanks"
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -625,7 +625,10 @@ class Recents(Module):
     SPEC = [
         ("dir", "AppData/Roaming/Microsoft/Windows/Recent", from_user_home),
         ("dir", "AppData/Roaming/Microsoft/Office/Recent", from_user_home),
+        ("glob", "AppData/Roaming/Microsoft/Windows/Start Menu/Programs/*.lnk", from_user_home),
+        ("glob", "Desktop/*.lnk", from_user_home),
         ("glob", "Recent/*.lnk", from_user_home),
+        ("glob", "sysvol/ProgramData/Microsoft/Windows/Start Menu/Programs/*.lnk"),
     ]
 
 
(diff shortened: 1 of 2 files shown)


Zawadidone:
The acquire container is larger compared to a Windows container, but these paths could contain valuable forensics traces and are needed by `target-query` (https://github.com/fox-it/dissect.target/blob/b6ea46e2ed95321440df6c8c6b65803af06da859/dissect/target/plugins/os/unix/bsd/osx/_os.py#L21) to detect the OSX operating system.

```bash
ls -lh dissect.sys_20230829104939.tar 
-rw-r--r-- 1 root staff 21G Aug 29 13:15 dissect.sys_20230829104939.tar
target-query -f hostname,os,version -t dissect.sys_20230829104939.tar  
<Target dissect.sys_20230829104939.tar> osx dissect.sys macOS 13.4 (22F66)
```

Schamper:
That's quite a bit larger. Can you give some examples of the forensic value? For the OS plugin, we can probably fix that some other way.

Perhaps it's an idea to only put this in the `full` profile for MacOS?

Schamper:
I see, that makes more sense!

acquire/acquire.py
```diff
@@ -1523,6 +1523,35 @@ class OSX(Module):
...
+class OSXApplications(Module):
+    DESC = "OS-X application files"
+    SPEC = [
+        # applications
+        ("glob", "/Applications/*/Contents/Info.plist"),
+        ("glob", "Applications/*/Contents/Info.plist", from_user_home),
+    ]
+
+    @classmethod
+    def get_spec_additions(cls, target):
```
Schamper, line 1536:
```suggestion
    def get_spec_additions(cls, target: Target) -> Iterator[tuple]:
```

acquire/acquire.py
```diff
@@ -1523,6 +1523,35 @@ class OSX(Module):
...
+        spec = set()
+
+        for appplication_support in itertools.chain(
+            ["/Library/Application Support"],
+            from_user_home(target, "Library/Application Support"),
+        ):
+            for path in target.fs.path(appplication_support).rglob("*"):
```
Schamper, line 1543:
```suggestion
        for app_support_path in itertools.chain(
            ["/Library/Application Support"],
            from_user_home(target, "Library/Application Support"),
        ):
            for path in target.fs.path(app_support_path).rglob("*"):
```

acquire/acquire.py
```diff
@@ -1523,6 +1523,35 @@ class OSX(Module):
     ]
 
 
+@register_module("--osx-applications")
+class OSXApplications(Module):
+    DESC = "OS-X application files"
+    SPEC = [
+        # applications
```
Schamper, line 1530:
```suggestion
        # Info.plist from all installed applications
```

acquire/acquire.py
```diff
@@ -1523,6 +1523,35 @@ class OSX(Module):
...
+        ("glob", "Applications/*/Contents/Info.plist", from_user_home),
+    ]
+
+    @classmethod
+    def get_spec_additions(cls, target):
+        spec = set()
+
+        for appplication_support in itertools.chain(
+            ["/Library/Application Support"],
+            from_user_home(target, "Library/Application Support"),
```
Schamper, line 1541: Are there more specific paths you can use instead of every file <=1MB from these paths? Doing a quick `find` on my personal Mac results in around 30GB of data, much of which are cache files and Steam game files 😉.

Also, don't these paths generally contain a lot of privacy sensitive information (session cookies and the like)? Whereas the other application paths contain just generic installed application information. Perhaps nice to separate them, include the `Info.plist` info in every profile (minimal, default and full), and a little bit more fleshed out Application Support module in full?

Schamper (changes requested):

acquire/acquire.py
Zawadidone: Added it to the description of the new module

acquire/acquire.py
Zawadidone: I have adjusted the pattern to only acquire files in directory and subdirectories of the applications, because these paths contain configuration files of the applications and are mostly used by attacker shows in the blog posts of Objective See.

The modules are seperated.

Poeloe: With the used `glob("*/*/*")` you're now searching for files exactly two sub-directories deep (no more, no less). From your comment, I think that's not the behaviour you wish for, is it?

If you indeed want to search more recursively, you might want to use `rglob("*")` or `rglob("*/*")` instead. This allows you to search for all files within the "Application Support" directory and/or all it's sub-directories.

Zawadidone: @Poeloe I should have said the files in the directory of the application and files in the sub-directory of an application, as shown below.

* `/Users/<USER>/Library/Application Support/`
* `/Users/<USER>/Library/Application Support/app1`
* `/Users/<USER>/Library/Application Support/app1/dir/`

When using rglob the files in the following directory are also collected, which I don't want.

`/Users/<USER>/Library/Application Support/app1/dir/dir/dir/dir/dir`

Poeloe: Maybe the solution is then to use `rglob` in combination with and if-statement like:

```python
if not path.is_file():
   continue

if path.parts[1] == "Library" and len(path.parts) > 5:
   continue

if len(path.parts) > 7:
   continue

if path.stat().st_size >= (1 * 1024 * 1024):  # 1MB
   continue
```

Does this give you the proper results?

Poeloe (changes requested):

acquire/acquire.py
```diff
@@ -1504,6 +1504,39 @@ class OSX(Module):
...
+        for app_support_path in itertools.chain(
+            ["/Library/Application Support"],
+            from_user_home(target, "Library/Application Support"),
+        ):
+            for path in target.fs.path(app_support_path).glob("*/*/*"):
+                if not path.is_file():
+                    continue
+
+                if path.stat().st_size >= (1 * 1024 * 1024):  # 1MB
+                    continue
```
Poeloe, line 1533: Hereby the suggestion code ;)

```suggestion
            for path in target.fs.path(app_support_path).rglob("*"):
                if not path.is_file():
                    continue
               
               if path.parts[1] == "Library" and len(path.parts) > 5:
                    continue

               if len(path.parts) > 7:
                    continue

               if path.stat().st_size >= (1 * 1024 * 1024):  # 1MB
                    continue
```

acquire/acquire.py
Zawadidone: The current implementation [`25467c4` (#96)](https://github.com/fox-it/acquire/pull/96/commits/25467c40c766dc17b18e330466fe396a22b71191) already gives the proper results. Is it needed to use `rglob` with your suggested code?

acquire/acquire.py
Zawadidone: The suggested code takes `3m49.462s` compared to `5.232s` [`ae3ec77` (#96)](https://github.com/fox-it/acquire/pull/96/commits/ae3ec772eb3895f8194527727f89035cfbd797f3), but the code looks a bit ugly.

```
[...]
   # Special case for the non-user directory "/Library/Application Support/"
                if path.parts[1] == "Library" and len(path.parts) > 6:
                    continue

                if len(path.parts) > 8:
                    continue
[...]
```

acquire/acquire.py
Zawadidone: That's true, I have removed the module.

How would acquire ASDF solve the issue of not having the luxury of a MFT for Unix based systems?

Schamper: Once ASDF support is included, you'll be able to completely browse any filesystem for which we have added ASDF support. Iterating files and inodes, browsing around with `target-shell` and `target-query` plugins  like `suidbinaries` will all work as expected, while still maintaining the small collection footprint that acquire has now. It's like having a full disk image, but smaller.

In addition, files that you have collected (e.g. the `Info.plist`) will be completely readable. Any file that you will not have collected, on the other hand, will return `\xa5\xdf` as its file content.

Poeloe (approved):

Schamper (approved):

PR #109, opened 2023-11-17
Provide PyOxidizer binaries

Schamper:
This PR adds CI for building PyOxidizer binaries for Windows, macOS, Linux (GNU) and Linux (musl). 

The Linux musl build currently depends on a forked PyOxidizer (https://github.com/fox-it/PyOxidizer/tree/esxi-compatibility) because of https://github.com/indygreg/PyOxidizer/pull/725. The fork also includes a workaround for ESXi.

The Linux GNU build is built in a `manylinux2014` container to ensure an old libc version for compatibility.

The macOS build combines a x86_64 and aarch64 binary into a single universal2 binary for compatibility on both Intel and Apple Silicon Macs.

A separate list of dependencies is currently maintained in `pyoxidizer.bzl` because there are no appropriate dependency bundles available in acquire right now. `minio` is included for uploading to cloud storage, but unfortunately `pycryptodome` is not yet included. It _should_ be possible to include it but I couldn't get it working across all OS variants right now, so that could be a future improvement. The way to go here is likely a filesystem fallback on an accompanied `lib` directory, but for some reason this broke the embedded `msgpack`. PyOxidizer should also be able to [recompile an extension into itself](https://gregoryszorc.com/docs/pyoxidizer/main/pyoxidizer_packaging_extension_modules.html#building-with-a-custom-distutils) but I did not look into this yet.

The `pyoxidizer.bzl` file also includes comments explaining how to build manually, as well as how you could modify the embedded configuration.

Ideally we add these binaries as files into the releases, but I wasn't sure what the desired CI trigger would be for that, so I've put it on tags for now. For this reason, the PyOxidizer also currently downloads the latest stable `acquire` from PyPI instead of copying from the local source repository.

### DIFF
--- a/.github/workflows/pyoxidizer.yml
+++ b/.github/workflows/pyoxidizer.yml
@@ -27,9 +27,8 @@ jobs:
             ~/.cargo/git/db/
           key: linux-cache
 
-      # We need target-query to build the plugin list, let acquire determine which version to pull in
       - name: Install dependencies
-        run: /opt/python/cp39-cp39/bin/pip install pyoxidizer acquire==${{ github.event.inputs.tag }}
+        run: /opt/python/cp39-cp39/bin/pip install pyoxidizer dissect
 
       - name: Build binary
         run: |
@@ -84,9 +83,8 @@ jobs:
       - name: Install forked PyOxidizer
         run: cargo install --git https://github.com/Schamper/pyoxidizer pyoxidizer --force --locked
 
-      # We need target-query to build the plugin list, let acquire determine which version to pull in
       - name: Install dependencies
-        run: /opt/python/cp39-cp39/bin/pip install acquire==${{ github.event.inputs.tag }}
+        run: /opt/python/cp39-cp39/bin/pip install dissect
 
       - name: Build binary
         run: |
@@ -123,9 +121,8 @@ jobs:
         with:
           python-version: '3.9'
 
-      # We need target-query to build the plugin list, let acquire determine which version to pull in
       - name: Install dependencies
-        run: pip install pyoxidizer acquire==${{ github.event.inputs.tag }}
+        run: pip install pyoxidizer dissect
 
       - name: Build binary
         run: |
@@ -162,9 +159,8 @@ jobs:
         with:
           python-version: '3.9'
 
-      # We need target-query to build the plugin list, let acquire determine which version to pull in
       - name: Install dependencies
-        run: pip install pyoxidizer acquire==${{ github.event.inputs.tag }}
+        run: pip install pyoxidizer dissect
 
       - name: Build binary
         run: |
--- a/.github/workflows/pyoxidizer.yml
+++ b/.github/workflows/pyoxidizer.yml
@@ -1,0 +1,180 @@
+name: Build PyOxidizer Binary
+on:
+  push:
+    tags:
+      - '*'
+  workflow_dispatch:
+
+jobs:
+  build-linux-gnu:
+    name: Build Linux (GNU)
+    runs-on: ubuntu-latest
+    container: quay.io/pypa/manylinux2014_x86_64
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pyoxidizer/
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: linux-cache
+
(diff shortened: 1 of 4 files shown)


.github/workflows/pyoxidizer.yml
```diff
@@ -0,0 +1,178 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pyoxidizer/
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: linux-cache
+
```
pyrco, line 26: Why don't you do a 
```
     - uses: actions/setup-python@v4
     ...
```
here? Saves on all the explicit paths.

.github/workflows/pyoxidizer.yml
```diff
@@ -0,0 +1,178 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pyoxidizer/
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: linux-musl-cache
+
```
pyrco, line 69: Same here?

pyrco (changes requested):

.github/workflows/pyoxidizer.yml
```diff
@@ -0,0 +1,178 @@
...
+          path: |
+            ~/.cache/pyoxidizer/
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: linux-cache
+
+      - name: Install dependencies
+        run: /opt/python/cp39-cp39/bin/pip install pyoxidizer dissect.target
```
pyrco, line 28: This runs the risk of installing dissect dev packages. It is better to install the `dissect` package.

However as this workflow triggers on tags, and acquire is usually tagged before a new dissect package is tagged&released, this will still not work. I think it is best to make this a manual workflow where a tag can be given as parameter.

.github/workflows/pyoxidizer.yml
Schamper: Unfortunately `actions/setup-python` depends on a newer LIBC and is incompatible with this container. Same for the other job.

.github/workflows/pyoxidizer.yml
Schamper: That's only when you add `--pre` right? This just installs the latest stable `dissect.target`.

.github/workflows/pyoxidizer.yml
pyrco: Maybe this as a comment on why it is not used.

.github/workflows/pyoxidizer.yml
pyrco: Yes, but it will only install the stable version of dissect.target not its dissect.* dependencies.

cuhsat:
In the meanwhile, I updated/fixed the CI from the original PR in my own repo, and were successful in creating binaries:

Miauwkeru (changes requested):

.github/workflows/pyoxidizer.yml
```diff
@@ -0,0 +1,183 @@
...
+
+      - name: Install dependencies
+        run: pip install pyoxidizer dissect
+
+      - name: Build binary
+        run: |
+          mkdir -p build/lib/dissect/target/plugins
+          target-build-pluginlist > build/lib/dissect/target/plugins/_pluginlist.py
+          pyoxidizer build --release --target-triple x86_64-pc-windows-msvc --var flavor standalone_static --var version ${{ github.event.inputs.tag }}
+          strip build/x86_64-pc-windows-msvc/release/install/acquire.exe
```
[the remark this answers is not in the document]
joost-j: Included it in the binary name

.github/workflows/pyoxidizer.yml
```diff
@@ -0,0 +1,183 @@
...
+
+      - name: Verify binary
+        run: build/x86_64-pc-windows-msvc/release/install/acquire.exe --version
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: acquire-windows
+          path: build/x86_64-pc-windows-msvc/release/install/*
```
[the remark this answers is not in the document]
joost-j: Not sure what you mean by this. But I'd say getting these pre-built binaries available to the public has priority right now, let's look into this later.

PR #112, opened 2023-12-18
Properly name special NTFS file in the collected output

pyrco:
(DIS-2897)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -328,8 +328,12 @@ class NTFS(Module):
                 journal.seek(journal.runlist[i][1] * journal.block_size, io.SEEK_CUR)
                 i += 1
 
+            # Use the same method to construct the output path as is used in
+            # collector.collect_file()
+            outpath = collector._output_path(f"{name}/$Extend/$Usnjrnl:$J")
+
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -328,8 +328,10 @@ def collect_usnjrnl(cls, collector: Collector, fs: Filesystem, name: str) -> Non
                 journal.seek(journal.runlist[i][1] * journal.block_size, io.SEEK_CUR)
                 i += 1
```
cecinestpasunepipe, line 330: Can you add a comment why this was done? Otherwise we might forget...

acquire/acquire.py
```diff
@@ -352,8 +354,11 @@ def collect_ntfs_secure(cls, collector: Collector, fs: Filesystem, name: str) ->
             secure_path = fs.path("$Secure:$SDS")
             entry = secure_path.get()
             sds = entry.open()
+
```
cecinestpasunepipe, line 357: Same here...

cecinestpasunepipe (changes requested):

cecinestpasunepipe (approved):

PR #119, opened 2024-01-17
Map all Windows sysvol paths to their drive letter

Schamper:
Required for https://github.com/fox-it/dissect.target/pull/497.

Allows `dissect.target` to support acquire outputs for systems that have a non-default sysvol drive letter.

### DIFF
--- a/acquire/utils.py
+++ b/acquire/utils.py
@@ -368,10 +368,13 @@ def persist_execution_report(path: Path, report_data: dict) -> Path:
         f.write(json.dumps(report_data, sort_keys=True, indent=4))
 
 
-SYSVOL_SUBST = re.compile(r"^(/\?\?/)?[cC]:")
+DEVICE_SUBST = re.compile(r"^(/\?\?/)")
+SYSVOL_SUBST = re.compile(r"^/?sysvol(?=/)")
 
 
-def normalize_path(target: Target, path: Path, resolve: bool = False, lower_case: bool = True) -> str:
+def normalize_path(
+    target: Target, path: Path, *, sysvol: Optional[str] = None, resolve: bool = False, lower_case: bool = True
+) -> str:
     if resolve:
         path = path.resolve()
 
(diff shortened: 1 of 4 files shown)


acquire/utils.py
```diff
@@ -381,7 +384,12 @@ def normalize_path(target: Target, path: Path, resolve: bool = False, lower_case
         path = path.lower()
 
     if target.os == "windows":
-        # As dissect.target always maps c: onto sysvol, we can do this substitution here.
-        path = SYSVOL_SUBST.sub("sysvol", path)
+        path = DEVICE_SUBST.sub("", path)
+        if sysvol:
+            path = normalize_sysvol(path, sysvol)
```
pyrco, line 389: I think it is better to put this before the if condition containing `path.lower()` so the drive letter always has a uniform case.

acquire/utils.py
```diff
@@ -368,10 +368,13 @@ def persist_execution_report(path: Path, report_data: dict) -> Path:
...
 
 
-SYSVOL_SUBST = re.compile(r"^(/\?\?/)?[cC]:")
+DEVICE_SUBST = re.compile(r"^(/\?\?/)")
+SYSVOL_SUBST = re.compile(r"^/?sysvol(?=/)")
 
 
-def normalize_path(target: Target, path: Path, resolve: bool = False, lower_case: bool = True) -> str:
+def normalize_path(
+    target: Target, path: Path, *, sysvol: Optional[str] = None, resolve: bool = False, lower_case: bool = True
```
pyrco, line 376: There is some ugly dependency between `normalize_path()` and `serialize_path()`. It seems the latter sometimes depends on paths already being put through the former.

Currently there are again duplicates being collected as `serialize_path()` does not know about the maping between sysvol and the drive letter. As `sysvol` is optional in `normalize_path` things like `from_user_home()` emit both sysvol and drive letter paths, which the dedup logic (using `serialize_path()`) does not catch.

Ideally we rewrite acquire to use single canonical paths, but maybe we can get away with not making sysvol optional, so it is replaced everywhere `normalize_path()` is used.

pyrco (changes requested):

acquire/utils.py
Schamper: Opted for the latter for now.

pyrco: Unfortunately it does not fix the de-duplication issue with `serialize_path()`, but for this PR I think that's fine as we've planned to tackle the de-dup issue separately.

pyrco (approved):

PR #122, opened 2024-01-23
Add additional SchedLgU.txt paths

Horofic:
(DIS-2136)


Poeloe (approved):
Approved! But when merging, please change the commit message, such that it starts with the imperative and put the DIS-* in the body ;)

PR #130, opened 2024-02-15
Bugfixes for esxi image and volatile-profiles

Miauwkeru:
- Fix volatile_profile issue with defaults enabled
- Fix issue when acquiring an esxi image

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -2193,7 +2193,7 @@ def main() -> None:
         log.exception("Failed to load target")
         raise
 
-    if target.os == "esxi":
+    if target.os == "esxi" and target.name == "local":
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -1768,8 +1768,12 @@ def acquire_target_regular(target: Target, args: argparse.Namespace, output_ts:
     profile_modules = _add_modules_for_profile(
         profile, target.os, PROFILES, "No collection set for OS %s with profile %s"
     )
+
+    if not (volatile_profile := args.volatile_profile):
+        volatile_profile = "none"
```
Schamper, line 1773: What is the benefit of this over the default value?

Schamper (approved):

PR #161, opened 2024-03-26
New collection and de-duplication logic

pyrco:
When paths contain intermediate symlinks, these intermediate symlinks are collected instead of collecting the unresolved path as is. De-duplication between sysvol and it's equivalent drive letter path is done properly.
Collecting of special files (e.g. MFT, Recyclebin etc.) is properly abstracted.

(DIS-1222)

### DIFF
--- a/acquire/collector.py
+++ b/acquire/collector.py
@@ -9,10 +9,9 @@ from collections import defaultdict
 from contextlib import contextmanager
 from dataclasses import dataclass
 from itertools import groupby
-from pathlib import Path
 from typing import (
     TYPE_CHECKING,
-    Any,
+    BinaryIO,
     Callable,
     Iterable,
     Optional,
@@ -28,14 +27,10 @@ from dissect.target.exceptions import (
     NotASymlinkError,
     SymlinkRecursionError,
 )
+from dissect.target.filesystem import Filesystem
 from dissect.target.helpers import fsutil
 
-from acquire.utils import (
-    StrEnum,
-    get_formatted_exception,
-    normalize_path,
-    normalize_sysvol,
-)
+from acquire.utils import StrEnum, get_formatted_exception, normalize_path
 
 if TYPE_CHECKING:
     from acquire.outputs.base import Output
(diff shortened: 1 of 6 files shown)


acquire/collector.py
```diff
@@ -289,114 +285,364 @@ def collect(
...
         if not module_name:
             raise ValueError("Module name must be provided or Collector needs to be bound to a module")
 
         if not isinstance(path, fsutil.TargetPath):
             path = self.target.fs.path(path)
 
-        if self.filter(path) is True:
-            log.info("- Collecting file %s: Skipped (filtered out)", path)
-            return
+        log.info("- Collecting path %s", path)
```
Schamper, line 361: With the new logging I get a lot of seemingly duplicate log lines printed to the console, can that be cleaned up more? Maybe log some things to debug (the log file) and only print the minimal amount of log lines necessary to the console (info).

For example, this is what I see for every single file:
```
- Collecting path /c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1
- Collecting file /c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1 to: fs/c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1
- Collecting file /c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1 succeeded
- Collecting path /c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1 succeeded
```

I liked the `... : OK` that we previously had, but I also like how now you're more easily able to track progress. Maybe something like this?
```
- Collecting file /c:/Users/User/AppData/Local/Microsoft/Windows/UsrClass.dat.LOG1 ...
- ... Done!
```

Maybe we can move towards `structlog` in the future to better facilitate this too.

acquire/collector.py
```diff
@@ -289,114 +285,364 @@ def collect(
...
+                return
 
-    def collect_symlink(self, path: fsutil.TargetPath, module_name: Optional[str] = None) -> None:
-        try:
-            outpath = self._output_path(path)
-            self.output.write_entry(outpath, path.get())
+            # If a path does not exist, is_dir(), is_file() and is_symlink() will return False (and
+            # not raise an exception), so we need to explicitly trigger an exception for this using
+            # collect_inpath.get().  This also checks if the path can be fully resolved.
+            path_entry = collect_inpath.get()
```
Schamper, line 417: `.stat()` is the more pathlib-like approach. Or `.exists()` ofcourse.

acquire/collector.py
pyrco: Yeah, that's a bit unfortunate. The initial collection will always be a 'path' and it is not clear upfront if this will result in a single file, directory or something else with or without a set of intermediate symlinks.

I don't really see a different way to indicate which set of artifacts belong to a certain path, while retaining the traceability where and which things fail to collect.

pyrco:
> My output also sometimes has mixed drive letter capitalization, notably for the PowerShell history and Startup LNK files. Ideally we normalize that too.

That is the same issue as generic path name uniformization. So see my comment about that.

Schamper (approved):

PR #165, opened 2024-04-04
Add AnyDesk paths

Poeloe:


acquire/acquire.py
```diff
@@ -1217,9 +1217,14 @@ class RemoteAccess(Module):
         ("glob", "sysvol/Program Files/TeamViewer/*.log"),
         ("glob", "sysvol/Program Files (x86)/TeamViewer/*.log"),
         ("glob", "AppData/Roaming/TeamViewer/*.log", from_user_home),
-        # anydesk
+        # anydesk - Windows
         ("dir", "sysvol/ProgramData/AnyDesk"),
         ("glob", "AppData/Roaming/AnyDesk/*.trace", from_user_home),
+        ("dir", "AppData/AnyDesk/", from_user_home),
+        ("dir", "AnyDesk/", from_user_home),
```
Schamper, line 1224: Are these correct?

Poeloe: Changed this (misunderstood the environment variables in the documentation) and changed from collecting entire directory to only the .trace files, such that we won't be surprised by the size of the other files that are present ;)

acquire/acquire.py
```diff
@@ -1217,9 +1217,13 @@ class RemoteAccess(Module):
         ("glob", "sysvol/Program Files/TeamViewer/*.log"),
         ("glob", "sysvol/Program Files (x86)/TeamViewer/*.log"),
         ("glob", "AppData/Roaming/TeamViewer/*.log", from_user_home),
-        # anydesk
+        # anydesk - Windows
         ("dir", "sysvol/ProgramData/AnyDesk"),
         ("glob", "AppData/Roaming/AnyDesk/*.trace", from_user_home),
+        ("glob", "AppData/Roaming/AnyDesk/*/*.trace", from_user_home),
```
Schamper, line 1223:
```suggestion
        ("glob", "AppData/Roaming/AnyDesk/**/*.trace", from_user_home),
```
Would this work?

acquire/acquire.py
```diff
@@ -1217,9 +1217,13 @@ class RemoteAccess(Module):
...
+        # anydesk - Mac + Linux
+        ("glob", ".anydesk*/*.trace", from_user_home),
```
Schamper, line 1225: Just checking - is it actually `.anydesk*` or should it be `.anydesk/*`?

Schamper (changes requested):

acquire/acquire.py
Poeloe: It should be like it is ;)

https://support.anydesk.com/knowledge/trace-files#trace-file-locations --> For the Custom Client, it could be located in `~/.anydesk_ad_<prefix>/anydesk_ad_<prefix>.trace` or for the Standard Client just `~/.anydesk/anydesk.trace`

acquire/acquire.py
Poeloe: It would work, but I don't know if it will negatively impact the runtime, since it will then traverse all of the paths

Schamper (approved):

PR #175, opened 2024-06-26
Fix shard issue

cecinestpasunepipe:
(DIS-3251)

### DIFF
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -44,6 +44,5 @@ def test_gui(
     GUI.__new__ = lambda x: Diagnostic_GUI()
     shards = []
     args = Namespace(child=False, auto_upload=auto_upload, children=children, skip_parent=skip_parent, start_time=0)
-    children = [mock_target] * children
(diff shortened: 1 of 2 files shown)


tests/test_gui.py
```diff
@@ -0,0 +1,49 @@
...
+        (50, False, True, list(range(0, 51))),  # Should not be zero filled...
+    ],
+)
+@patch("acquire.gui.base.Stub", spec=True)
+@patch("acquire.acquire.acquire_target", create=True)
+def test_gui(
+    mock_target: Target, gui: GUI, children: int, skip_parent: bool, auto_upload: bool, expected_shards: list[int]
+):
+    def list_children():
+        return children
```
pyrco, line 31:
```suggestion
        yield from [mock_target] * num_children
```

I think this way it is more clear what you are meaning to do, and it is a bit more in line with the actual list_children functions being iterators.

pyrco (changes requested):

pyrco (approved):

PR #182, opened 2024-07-08
Fix compression in zip outputs when compress option is supplied

ruzzle:
The zip outputs of acquire ignored the compression type given, due to the use of zipfile.ZipInfo without supplying a compression type to these objects. This PR fixes that.

I also changed the compression type to deflate, as this is better supported throughout other forensic tools (read: tools we use ourselves to process acquire collects)

### DIFF
--- a/acquire/outputs/zip.py
+++ b/acquire/outputs/zip.py
@@ -38,7 +38,7 @@ class ZipOutput(Output):
         self.path = path.with_suffix(path.suffix + ext)
 
         if compress:
-            self.compression = zipfile.ZIP_LZMA
+            self.compression = zipfile.ZIP_DEFLATED
         else:
             self.compression = zipfile.ZIP_STORED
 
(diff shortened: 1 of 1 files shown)


ruzzle:
@DissectBot agree [company="Northwave"]

Horofic:
Nice catch @ruzzle! Ideally this should be something configurable, though that is out of the scope for this PR. Would you be able to say which tools currently have trouble reading LZMA compressed ZIP files?

ruzzle:
Specifically, we are using velociraptor to process dissect acquire collects via its deaddisk remapping functionality. See https://docs.velociraptor.app/blog/2022/2022-03-22-deaddisk/

We try to tune acquire to output compatible output (zip / deflate), velociraptor doesn't like lzma apparently

Schamper:
Is there already a bug report for Velo to support LZMA? Besides us supporting one or the other, I think Velo could also benefit.

ruzzle:
> Is there already a bug report for Velo to support LZMA? Besides us supporting one or the other, I think Velo could also benefit.

Now there is 😉: https://github.com/Velocidex/velociraptor/issues/3608

Horofic (approved):

PR #185, opened 2024-07-10
Make acquire compression configurable

Horofic:
Make `tar` and `zip` compression output configurable by supplying a `--compress-method` argument as mentioned in #182 

Additionally, some tests have been adjusted to account for this change.

### DIFF
--- a/acquire/utils.py
+++ b/acquire/utils.py
@@ -328,14 +328,15 @@ def check_and_set_acquire_args(
     if not args.children and args.skip_parent:
         raise ValueError("--skip-parent can only be set with --children")
 
-    if args.output_type == "zip" and args.compress_method not in ZIP_COMPRESSION_METHODS:
-        raise ValueError(
-            f"Invalid compression method for zip, allowed are: {', '.join(ZIP_COMPRESSION_METHODS.keys())}"
-        )
-    if args.output_type == "tar" and args.compress_method not in TAR_COMPRESSION_METHODS:
-        raise ValueError(
-            f"Invalid compression method for tar, allowed are: {', '.join(TAR_COMPRESSION_METHODS.keys())}"
-        )
+    if args.compress:
+        if (args.output_type == "zip" and args.compress_method) and args.compress_method not in ZIP_COMPRESSION_METHODS:
+            raise ValueError(
+                f"Invalid compression method for zip, allowed are: {', '.join(ZIP_COMPRESSION_METHODS.keys())}"
+            )
+        if (args.output_type == "tar" and args.compress_method) and args.compress_method not in TAR_COMPRESSION_METHODS:
+            raise ValueError(
+                f"Invalid compression method for tar, allowed are: {', '.join(TAR_COMPRESSION_METHODS.keys())}"
+            )
 
 
 def get_user_name() -> str:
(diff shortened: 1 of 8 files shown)


acquire/utils.py
```diff
@@ -84,6 +84,11 @@ def create_argument_parser(profiles: dict, volatile: dict, modules: dict) -> arg
         action="store_true",
         help="compress output (if supported by the output type)",
     )
+    parser.add_argument(
+        "--compress-method",
+        choices=COMPRESSION_METHODS,
```
pyrco, line 89: I think it would be good if `check_and_set_acquire_args()` would check if the given compression method is allowed for the given compression type.

pyrco (changes requested):

acquire/utils.py
```diff
@@ -323,6 +328,11 @@ def check_and_set_acquire_args(
     if not args.children and args.skip_parent:
         raise ValueError("--skip-parent can only be set with --children")
 
+    if args.output_type == "zip" and args.compress_method not in ZIP_COMPRESSION_METHODS:
+        raise ValueError(f"Invalid compression method for zip: {ZIP_COMPRESSION_METHODS.keys()}")
```
pyrco, line 332:
```suggestion
        raise ValueError(f"Invalid compression method for zip, allowed are: {ZIP_COMPRESSION_METHODS.keys()}")
```

Otherwise it reads as if these methods are invalid

acquire/utils.py
```diff
@@ -323,6 +328,11 @@ def check_and_set_acquire_args(
...
+    if args.output_type == "tar" and args.compress_method not in TAR_COMPRESSION_METHODS:
+        raise ValueError(f"Invalid compression method for tar: {TAR_COMPRESSION_METHODS.keys()}")
```
pyrco, line 334:
```suggestion
        raise ValueError(f"Invalid compression method for tar, allowed are: {TAR_COMPRESSION_METHODS.keys()}")
```

pyrco (changes requested):

pyrco (changes requested):
LGTM, but the testcases now fail

acquire/utils.py
```diff
@@ -318,6 +328,15 @@ def check_and_set_acquire_args(
...
 
+    if args.output_type == "zip" and args.compress_method not in ZIP_COMPRESSION_METHODS:
+        raise ValueError(
+            f"Invalid compression method for zip, allowed are: {', '.join(ZIP_COMPRESSION_METHODS.keys())}"
+        )
+    if args.output_type == "tar" and args.compress_method not in TAR_COMPRESSION_METHODS:
+        raise ValueError(
+            f"Invalid compression method for tar, allowed are: {', '.join(TAR_COMPRESSION_METHODS.keys())}"
+        )
+
```
pyrco, line 339:
```suggestion
    if args.compress:
        if args.output_type == "zip" and args.compress_method not in ZIP_COMPRESSION_METHODS:
            raise ValueError(
                f"Invalid compression method for zip, allowed are: {', '.join(ZIP_COMPRESSION_METHODS.keys())}"
            )
        if args.output_type == "tar" and args.compress_method not in TAR_COMPRESSION_METHODS:
            raise ValueError(
                f"Invalid compression method for tar, allowed are: {', '.join(TAR_COMPRESSION_METHODS.keys())}"
            )

```

This should fix  the test cases I think. I don't know if you can have a zip output without compression though

pyrco (changes requested):

acquire/utils.py
pyrco: It looks like that is possible.

pyrco (approved):

pyrco (approved):

PR #186, opened 2024-07-12
Add exit codes and summary to acquire-decrypt

Poeloe:


acquire/tools/decrypter.py
```diff
@@ -265,51 +265,64 @@ def check_existing(in_path: Path, out_path: Path, status_queue: multiprocessing.
...
-                    _info(status_queue, f"{in_path} • Verify error: {e}")
-                    _info(status_queue, f"{in_path} • ! Output file should NOT be trusted")
-                except Exception as e:
-                    _info(status_queue, f"{in_path} • Unknown error: {e}")
+            except ValueError as e:
+                message = f"Error decrypting file: {e}"
+                _info(status_queue, f"{in_path} • {message}")
+            except VerifyError as e:
+                _info(status_queue, f"{in_path} • Verify error: {e}")
+                message = "! Output file should NOT be trusted"
```
pyrco, line 310: Maybe add something like `Verification failed` to the message, so it is clear why the output should not be trusted.

acquire/tools/decrypter.py
```diff
@@ -444,6 +461,22 @@ def main():
                     executor.shutdown(wait=True, cancel_futures=True)
                     break
 
+            if len(files) > 1:
```
pyrco, line 464: Why only for more then 1 file?

acquire/tools/decrypter.py
```diff
@@ -444,6 +461,22 @@ def main():
...
+                        f"{file} - {'Success' if success else 'Failed'}: {message}"
+                        for file, (success, message) in sorted(results.items())
+                    ),
+                    " • ",
+                )
+                (progress.console.log if progress else log.info)(
+                    f"\nDecrypt results (file - result: message):\n{results_string}"
+                )
+
+            if not all(results.values()) or tasks:
```
pyrco, line 476: If there are tasks left in the queue, do they have a result output? Should this case not be handled separately?

acquire/tools/decrypter.py
```diff
@@ -444,6 +461,22 @@ def main():
...
+                exit_code = 2 if any(results.values()) else 1
```
pyrco, line 477: You probably only want to check on the 1st item in `results.values()` otherwise the `any(...)` will always return `True`.

pyrco (changes requested):

acquire/tools/decrypter.py
Poeloe: I thought it would be a bit overkill to show a summary when you're only targeting one file.

So, happy to show the results also for the one file case

acquire/tools/decrypter.py
Poeloe: This is needed for the exit code. Because I could think of a case where all the files that were processed were successful, but there are still tasks that needed to be processed. In that case, the run was still partly successful and therefore we need to return a 2 exit code (2, because any() will then always evaluate to True)

acquire/tools/decrypter.py
Poeloe: True! Oversight after refactor :)

acquire/tools/decrypter.py
pyrco: It would make the output more uniform ;-)

acquire/tools/decrypter.py
```diff
@@ -444,6 +461,21 @@ def main():
...
+                    f"{file} - {'Success' if success else 'Failed'}: {message}"
+                    for file, (success, message) in sorted(results.items())
+                ),
+                " • ",
+            )
+            (progress.console.log if progress else log.info)(
+                f"\nDecrypt results (file - result: message):\n{results_string}"
+            )
+
+            if not all(successes := (success for success, _ in results.values())) or tasks:
```
pyrco, line 475: `successes` will not be what you think it is. Depending on the order of `True`'s and `False`s, it will either be `[]` or `[True]`.

Just move it out of `all()`

pyrco (changes requested):

acquire/tools/decrypter.py
pyrco: I see that it would return:

- 1 if all processes fail, irrespective if there are tasks left
- 1 if there are no results, irrespective if there are tasks left
- 2 if at least 1 process succeeds and the others fail, irrespective if there are tasks left
- 2 if all processes succeed but there are tasks left

So complete fail == 1, partial success == 2.

It is difficult to see from the code that this is the result. So it should get a comment, but better rewrite it to be a bit easier to parse.

Poeloe: I've rewritten it and added some comments

pyrco (approved):

PR #198, opened 2024-10-22
Acquire device list

cecinestpasunepipe:
(DIS-897)

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import argparse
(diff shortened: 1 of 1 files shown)


Horofic (commented):
It might be a good idea to use the `_windows_get_devices()` and co functions in dissect.target's local loader (`loader.py`) [https://github.com/fox-it/dissect.target/blob/main/dissect/target/loaders/local.py#L337C1-L337C28](url)

Poeloe (approved):
LGTM

PR #199, opened 2024-11-01
Add collection of Docker logs

Paradoxis:
Added docker logs, hostname and container configs to acquire and included it in the `FULL` Linux profile

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1972,6 +1972,7 @@ class WindowsProfile:
         SSH,
         IIS,
         TextEditor,
+        Docker,
     ]
 
 
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1372,6 +1372,15 @@ class SSH(Module):
             super().run(target, cli_args, collector)
 
 
+@register_module("--docker")
+class Docker(Module):
+    SPEC = [
+        ("glob", "/var/lib/docker/containers/*/*-json.log"),
+        ("glob", "/var/lib/docker/containers/*/*.json"),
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -1375,9 +1375,19 @@ def run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector)
 @register_module("--docker")
 class Docker(Module):
     SPEC = [
```
Horofic, line 1377: Could you add a description (`DESC`), like the other modules.

Horofic (changes requested):

acquire/acquire.py
```diff
@@ -1966,6 +1985,7 @@ class LinuxProfile:
     DEFAULT = MINIMAL
     FULL = [
         *DEFAULT,
+        Docker,
```
Horofic, line 1988: This can should also be added to the `WindowsProfile`

acquire/acquire.py
```diff
@@ -1372,6 +1372,25 @@ def run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector)
             super().run(target, cli_args, collector)
 
 
+@register_module("--docker")
+class Docker(Module):
```
Schamper, line 1376:
```suggestion
class Docker(Module):
    DESC = "various Docker logs and configuration files"
```

Schamper (changes requested):

Horofic (approved):

Schamper (approved):

PR #229, opened 2025-04-02
Add Applications and Network modules

### LINKED ISSUE #227: Missing collection paths
Looking at what dissect can parse I stumbled across two occurrences where `acquire.py` doesn't collect something that `target-query` can parse:

1. https://github.com/fox-it/dissect.target/blob/bb9a48852a33e58ea760d7901fd5170526076498/dissect/target/plugins/os/unix/linux/network.py#L205
```
/usr/lib/systemd/network
```

2. https://github.com/fox-it/dissect.target/blob/bb9a48852a33e58ea760d7901fd5170526076498/dissect/target/plugins/os/unix/applications.py#L15
```
/usr/share/applications
```

Looks like there is no glob / dir statement for `acquire.py` to collect these files although `target-query` can parse it.

This probably is true for more items, but it's quite complex to find a good mapping between paths in the `dissect.target` project and the `acquire` project.

OlafHaalstra:
Add support for `network` and `applications` module from `dissect.target`.

closes #227

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -86,9 +86,7 @@ ACQUIRE_BANNER = rf"""
  \__,_|\___\__, |\__,_|_|_|  \___|
   by Fox-IT   |_|             v{VERSION}
   part of NCC Group
-"""[
-    1:
-]
+"""[1:]
 
 MODULES = {}
 MODULE_LOOKUP = {}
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -1460,6 +1460,27 @@ class BSD(Module):
...
     )
 
+@register_module("--applications")
+class Applications(Module):
+    SPEC = (
+        ("dir", "/usr/share/applications")
+        ("dir", "/usr/local/share/applications")
+        ("dir", "/var/lib/snapd/desktop/applications")
+        ("dir", "/var/lib/flatpak/exports/share/applications")
+        ("dir", ".local/share/applications", from_user_home)
```
Horofic, line 1470:
```suggestion
        ("dir", "/usr/share/applications"),
        ("dir", "/usr/local/share/applications"),
        ("dir", "/var/lib/snapd/desktop/applications"),
        ("dir", "/var/lib/flatpak/exports/share/applications"),
        ("dir", ".local/share/applications", from_user_home),
```

Horofic:
@OlafHaalstra would you be able to make the suggested changes? Else, I could commit the changes and merge.

Horofic (approved):

Horofic (approved):

PR #238, opened 2025-04-17
Add GUI support for selecting files to collect/upload

### LINKED ISSUE #252: Add GUI support for selecting files to collect/upload
This issue is to enable follow up on #238

raverburg:
This update adds functionality to the GUI that allows users to select specific files for collection and/or upload. Selected files are passed to the existing logic used by acquire, allowing users to either collect them during acquisition or upload them independently.
 
Notes:
We use a custom version of acquire internally that looks for a config.json file in the same directory instead of using config.py


Miauwkeru (changes requested):

raverburg:
Hi! Sorry for the long wait. I only remembered this PR was still open when I ran into the same issue again: there's no easy way for clients to upload investigation material (without 3rd party tools), which is exactly what acquire GUI aims to fix.

I've addressed your feedback:
 - Correct ctypes pointer types on OPENFILENAME fields
 - Declare argtypes/restype for CommDlgExtendedError
 - Fix nMaxFile to use character count instead of byte size
 - Surface FNERR_BUFFERTOOSMALL error via MessageBox
 - Disable "Choose files" button once acquisition starts                                                                                                                                                                                                
 - Update wait_for_start return type to reflect list[str] for files

acquire/gui/base.py
```diff
@@ -92,8 +93,8 @@ class Stub(GUI):
     def message(self, message: str) -> None:
         pass
 
-    def wait_for_start(self, args: Namespace) -> tuple[str, bool, bool]:
-        return args.output, args.auto_upload, False
+    def wait_for_start(self, args: Namespace) -> tuple[str, list[str], bool, bool]:
+        return args.output, None, args.auto_upload, False
```
Miauwkeru:
```suggestion
        return args.output, [], args.auto_upload, False
```
Either an empty list or change the typehint that you know you could get a `None` value

acquire/gui/win32.py
```diff
@@ -310,11 +357,65 @@ def choose_folder(self) -> None:
...
+            if selected_paths:
+                self.files = [path for path in selected_paths if path]
+                user32.SetWindowTextA(self.file_label, f"{len(self.files)} file(s) selected".encode())
+                user32.SetWindowTextA(self.choose_files_button, b"Clear file(s)")
+                user32.EnableWindow(self.start_button, True)
+        else:
+            error_code = comdlg32.CommDlgExtendedError()
+            if error_code == FNERR_BUFFERTOOSMALL:
+                user32.MessageBoxA(
+                    self.hwnd, b"Too many or too long file names selected. Please select fewer files.", b"Acquire", 0
```
Miauwkeru:
```suggestion
                    self.hwnd, b"The combined filename length exceeded the expected size. Please select fewer files.", b"Acquire", 0
```
This reads better to me, but feel free to add some suggestions

Miauwkeru (changes requested):
If acquire output folder is selected, it might be good to disable the 
"select file(s) to collect" chooser. 
As currently you can select both where the upload part will get ignored. Looking just at the GUI you would not expect that to happen

acquire/gui/win32.py
raverburg: No I agree

acquire/gui/win32.py
```diff
@@ -310,11 +358,71 @@ def choose_folder(self) -> None:
...
+        buffer = create_unicode_buffer(buffer_size)
+
+        ofn = OPENFILENAME()
+        ofn.lStructSize = sizeof(OPENFILENAME)
+        ofn.hwndOwner = self.hwnd
+        ofn.lpstrFile = cast(buffer, w.LPWSTR)
+        ofn.nMaxFile = buffer_size
+        ofn.lpstrFilter = "All Files\0*.*\0Text Files\0*.txt\0\0"
+        ofn.nFilterIndex = 1
+        ofn.Flags = 0x00002000 | 0x00080000 | 0x00000200
```
Miauwkeru: Could you add a comment with what these magic numbers actually mean?

raverburg: I noticed I was using a wrong flag so I corrected this and changed the numbers with named constants

acquire/gui/win32.py
```diff
@@ -144,6 +151,40 @@ class ITEMIDLIST(Structure):
     _fields_ = (("mkid", SHITEMID),)
 
 
+class OPENFILENAME(Structure):
```
Miauwkeru:
```suggestion
class OPENFILENAMEW(Structure):
```
Just a rename of this symbol to indicate which structure it mirrors (cause there is also an `A` variant)

Miauwkeru (changes requested):

acquire/gui/win32.py
Miauwkeru: I haven't checked the other structures yet, but if these use a `W` too, maybe a good idea to add them to reduce ambiguity

acquire/gui/win32.py
```diff
@@ -144,6 +151,40 @@ class ITEMIDLIST(Structure):
...
+        ("lpfnHook", w.LPVOID),
+        ("lpTemplateName", w.LPCSTR),
+        ("pvReserved", w.LPVOID),
+        ("dwReserved", w.DWORD),
+        ("FlagsEx", w.DWORD),
+    )
+
+
+comdlg32 = WinDLL("comdlg32", use_last_error=True)
+comdlg32.GetOpenFileNameW.argtypes = (POINTER(OPENFILENAME),)
```
Miauwkeru:
```suggestion
comdlg32.GetOpenFileNameW.argtypes = (POINTER(OPENFILENAMEW),)
```

acquire/gui/win32.py
```diff
@@ -310,11 +363,71 @@ def choose_folder(self) -> None:
...
+            user32.EnableWindow(self.choose_folder_button, True)
+            if not self.folder:
+                user32.EnableWindow(self.start_button, False)
+            return
+
+        buffer_size = 4096
+        buffer = create_unicode_buffer(buffer_size)
+
+        ofn = OPENFILENAME()
+        ofn.lStructSize = sizeof(OPENFILENAME)
```
Miauwkeru:
```suggestion
        ofn = OPENFILENAMEW()
        ofn.lStructSize = sizeof(OPENFILENAMEW)
```

Miauwkeru (commented):
For consistency :P

Miauwkeru:
I reordered some function definitions so its a bit more visually distinct what belongs together. This was some general cleanup that wasn't related to the PR, so I did it myself. I will now do some final checks!

Miauwkeru (approved):

raverburg:
Great! Thank you for everything @Miauwkeru :smile:

Miauwkeru:
no problem @raverburg :)

PR #240, opened 2025-05-01
List a targets sub-targets/children from acquire using --list-children

### LINKED ISSUE #239: List a targets sub-targets/children with Acquire (Improvement/Ease of use)
Acquire is able to collect information from sub-targets (VMs) with the --children flag.  It is also possible to specify a single child with --child. However for the various hypervisor platforms on which this is applicable the means to list and identify the right file to supply to --child is not straightforward. 

Because Acquire already is able to identify a targets children, why not just easy expose these in a human friendly manner with a --list-children argument. 

I will create a PR for this.

lhaagsma:
Added a new command line flag "--list-children" that from acquire's main function only returns a human readable list of child targets and exits. 

Aims to close #239


acquire/acquire.py
```diff
@@ -2251,6 +2251,12 @@ def main() -> None:
                 target_name = target.name
                 log.info("Loading target %s", target_name)
                 log.info(target)
+                # Check if we are just listing children and not collecting
+                if args.list_children:
+                    log.info(f"Listing children on Target: %s", target.name)
```
Schamper, line 2256:
```suggestion
                    log.info("Listing children on target: %s", target_name)
```

acquire/acquire.py
```diff
@@ -2251,6 +2251,12 @@ def main() -> None:
...
+                    for index, child in enumerate(target.list_children()):
+                        log.info(f"- Found child #%s: type=%s, path=%s", index, child.type, child.path)
```
Schamper, line 2258:
```suggestion
                        log.info("- [#%s]: type=%s, path=%s", index, child.type, child.path)
```
Dunno if this looks any nicer?

Schamper (changes requested):
This would also be a nice addition to `target-query`, just saying 😉.

acquire/acquire.py
```diff
@@ -2419,6 +2419,12 @@ def main() -> None:
                 target_name = target.name
                 log.info("Loading target %s", target_name)
                 log.info(target)
+                # Check if we are just listing children and not collecting
+                if args.list_children:
+                    log.info(f"Listing children on Target: %s", target.name)
```
Schamper, line 2424:
```suggestion
                    log.info("Listing children on Target: %s", target.name)
```

acquire/acquire.py
```diff
@@ -2419,6 +2419,12 @@ def main() -> None:
...
+                    for index, child in enumerate(target.list_children()):
+                        log.info(f"- Found child #%s: type=%s, path=%s", index, child.type, child.path)
```
Schamper, line 2426:
```suggestion
                        log.info("- Found child #%s: type=%s, path=%s", index, child.type, child.path)
```

Schamper: Also maybe just make this look similar to how `target-query` shows it?

Schamper (changes requested):

lhaagsma:
Reworked this PR to use the logic added in target. 

`from dissect.target.tools.utils import _OverrideRequiredAction, list_children`

By importing these two functions it works the same. 

Alternatively the `_OverrideRequiredAction` could be copied from `target/tools/utils.py` to `acquire/utils.py`. But this feels like duplicating code.

PR #253, opened 2025-07-15
Add Splashtop paths

### LINKED ISSUE #152: Add Atera/Splashtop to Acquire
During a CERT case it was observed that the actors were using the Atera Management Agent. This agent seems to use the Splashtop Remote Access Tool underlying. We'll need to add these locations to acquire so we can query this data with target-query.

File locations: C:\Program Files (x86)\Splashtop\Splashtop Remote\Server\log\
 * svcinfo.txt -> Splashtop service information loggin;
 * agent_log.txt -> agent output, generic information;
 * sysinfo.txt -> information about server and session startups;
 * SPLog.00x -> information about clipboard, transferred files, etc;

bobkarreman:
closes #152

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -1333,6 +1333,9 @@ class RemoteAccess(Module):
         ("path", "sysvol/ProgramData/TightVNC/Server/Logs"),
         # Remote desktop cache files
         ("path", "AppData/Local/Microsoft/Terminal Server Client/Cache", from_user_home),
+        # Splashtop
+        ("path", "sysvol/ProgramData/Splashtop/Temp/log"),
+        ("path", "sysvol/Program Files (x86)/Splashtop/Splashtop Remote/Server/log"),
     )
 
 


acquire/acquire.py
```diff
@@ -1333,6 +1333,9 @@ class RemoteAccess(Module):
         ("path", "sysvol/ProgramData/TightVNC/Server/Logs"),
         # Remote desktop cache files
         ("path", "AppData/Local/Microsoft/Terminal Server Client/Cache", from_user_home),
+        # Splashtop
+        ("path", "sysvol/ProgramData/Splashtop/Temp/log"),
```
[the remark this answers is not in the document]
bobkarreman: Added a comment in the plugin. 

The folder contains a file transfer specific log which from what I can tell also included in the generic log. But since those are interesting events and the log would probably not be that big I included it just to be sure all logging about file transfers is captured.

acquire/acquire.py
```diff
@@ -1333,6 +1333,9 @@ class RemoteAccess(Module):
...
+        ("path", "sysvol/Program Files (x86)/Splashtop/Splashtop Remote/Server/log"),
```
Schamper, line 1338: They only have a 32bit version of the application?

bobkarreman: As far as I can find yes

PR #257, opened 2025-07-28
Add SharePoint Server logs

respondersGY:
These logs are very useful to detect ToolShell exploitation events

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -871,7 +871,7 @@ class SharePoint(Module):
         for reg_key in target.registry.glob_ext(key):
             try:
                 spec.add(("path", reg_key.value("LogDir").value))
-            except Exception:
+            except Exception:  # noqa: PERF203
                 pass
 
         return spec
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -858,6 +858,20 @@ class IIS(Module):
         spec.update(("path", log_path) for log_path in chain(*iis_plugin.log_dirs.values()))
         return spec
 
+@register_module("--spse")
+class SPSE(Module):
+    DESC = "Windows Share Point Server (SPSE) logs"
+
+    @classmethod
+    def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Iterator[tuple]:
+        spec = set()
+        key = "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Shared Tools\\Web Server Extensions\\16.0\\WSS"
+
+        for reg_key in target.registry.keys(key):
+            spec.add(("path", reg_key.value("LogDir").value))
+
+        return spec
+
 
 @register_module("--prefetch")
 class Prefetch(Module):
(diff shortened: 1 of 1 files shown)


acquire/acquire.py
```diff
@@ -858,6 +858,20 @@ def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Ite
...
         return spec
 
+@register_module("--spse")
+class SPSE(Module):
+    DESC = "Windows Share Point Server (SPSE) logs"
+
+    @classmethod
+    def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Iterator[tuple]:
+        spec = set()
+        key = "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Shared Tools\\Web Server Extensions\\16.0\\WSS"
```
Schamper, line 868: Can you glob or iterate over the version subkey? Can use the Exchange or MSSQL module as example.

Also linter is probably going to complain about only one newline above this.

Schamper (changes requested):

acquire/acquire.py
```diff
@@ -859,6 +859,24 @@ def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Ite
         return spec
 
 
+@register_module("--spse")
+class SPSE(Module):
+    DESC = "Windows Share Point Server (SPSE) logs"
```
Schamper, line 864:
```suggestion
    DESC = "Windows SharePoint Server (SPSE) logs"
```

acquire/acquire.py
```diff
@@ -859,6 +859,24 @@ def get_spec_additions(cls, target: Target, cli_args: argparse.Namespace) -> Ite
         return spec
 
 
+@register_module("--spse")
+class SPSE(Module):
```
Schamper, line 863: I can't really find much resources online referring to this as "SPSE", maybe just use `--sharepoint`? We also do `--exchange`.

acquire/acquire.py
Schamper: Yes. I assume you've tested this? Don't feel like spinning up a whole SharePoint system 😄.

Schamper:
> The linter fails, should the try-except be removed? 
> 
> 
> 
> ```
> 
> fix: commands[1]> ruff check --fix acquire tests
> 
> acquire/acquire.py:874:13: PERF203 `try`-`except` within a loop incurs performance overhead
> 
>     |
> 
> 872 |               try:
> 
> 873 |                   spec.add(("path", reg_key.value("LogDir").value))
> 
> 874 | /             except Exception:
> 
> 875 | |                 pass
> 
>     | |____________________^ PERF203
> 
> 876 |
> 
> 877 |           return spec
> 
>     |
> 
> 
> 
> Found 1 error.
> 
> ```

No, you can add a `# noqa: PERF203` comment on that line. I'm not sure on the exact syntax but the Ruff VSCode extension can do that automatically as an IntelliSense fix.

Schamper (approved):

PR #261, opened 2025-08-12
Prevent .tar file corruption by patching short reads

joost-j:
Fixes an issue where a `.tar` output file would contain inconsistencies with regards to expected and actual file size of the included files.

In some cases, a file on disk can report a size of X bytes, but at the time of actually reading X bytes from the file, less than X bytes are actually available in the file (a short read). Acquire would report these issues as `OSError` in the resulting Acquisition log file, because the Python stdlib `tarfile.py` handles it that way. Data may however already be written to the destination archive at that point. 

Afterwards, Acquire continues to add new files to the archive. When trying to untar the file using `tar -xvf <FILE>` this would show as a `tar: Skipping to next header` error and finally, the process exists with a nonzero exit code.

Included a test case which simulates a file that actually returns less bytes than its reported size, to test this case.

### DIFF
--- a/acquire/outputs/tar.py
+++ b/acquire/outputs/tar.py
@@ -4,6 +4,7 @@ import io
 import shutil
 import tarfile
 from typing import TYPE_CHECKING, BinaryIO
+from unittest.mock import patch
 
 from acquire.crypt import EncryptedStream
 from acquire.outputs.base import Output
@@ -34,6 +35,7 @@ def copyfileobj(
 
     blocks, remainder = divmod(length, bufsize)
     for _ in range(blocks):
+        # Already prevents "long reads" because it reads at max bufsize bytes at a time
         buf = src.read(bufsize)
         if len(buf) < bufsize:
             # raise exception("unexpected end of data")
@@ -42,6 +44,7 @@ def copyfileobj(
         dst.write(buf)
 
     if remainder != 0:
+        # Already prevents "long reads" because it reads at max bufsize bytes at a time
         buf = src.read(remainder)
         if len(buf) < remainder:
             # raise exception("unexpected end of data")
@@ -49,9 +52,6 @@ def copyfileobj(
             buf += b"\x00" * (remainder - len(buf))
         dst.write(buf)
     return
-
-
-tarfile.copyfileobj = copyfileobj
 
 
 class TarOutput(Output):
@@ -139,7 +139,8 @@ class TarOutput(Output):
             if stat:
                 info.mtime = stat.st_mtime
 
-        self.tar.addfile(info, fh)
+        with patch("tarfile.copyfileobj", copyfileobj):
+            self.tar.addfile(info, fh)
 
     def close(self) -> None:
         """Closes the tar file."""
--- a/acquire/outputs/tar.py
+++ b/acquire/outputs/tar.py
@@ -129,7 +129,6 @@ class TarOutput(Output):
                 # Prevents "long reads" because it reads at max bufsize bytes at a time
                 buf = fh.read(bufsize)
                 if len(buf) < bufsize:
-                    # raise exception("unexpected end of data")
                     # PATCH; instead of raising an exception, pad the data to the desired length
                     buf += tarfile.NUL * (bufsize - len(buf))
                 self.tar.fileobj.write(buf)
@@ -138,7 +137,6 @@ class TarOutput(Output):
                 # Prevents "long reads" because it reads at max bufsize bytes at a time
                 buf = fh.read(remainder)
                 if len(buf) < remainder:
-                    # raise exception("unexpected end of data")
                     # PATCH; instead of raising an exception, pad the data to the desired length
                     buf += tarfile.NUL * (remainder - len(buf))
                 self.tar.fileobj.write(buf)
--- a/acquire/outputs/tar.py
+++ b/acquire/outputs/tar.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import io
+import shutil
 import tarfile
 from typing import TYPE_CHECKING, BinaryIO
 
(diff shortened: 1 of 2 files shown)


Schamper (changes requested):

acquire/outputs/tar.py
```diff
@@ -15,6 +16,44 @@
...
+        buf = src.read(remainder)
+        if len(buf) < remainder:
+            # raise exception("unexpected end of data")
+            # PATCH; instead of raising an exception, pad the data to the desired length
+            buf += b"\x00" * (remainder - len(buf))
+        dst.write(buf)
+    return
+
+
+tarfile.copyfileobj = copyfileobj
```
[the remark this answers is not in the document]
joost-j: Replaced it with `unittest.mock`'s patch function. Despite that it is used mostly while testing, it fits this purpose imo.

tests/test_outputs_tar.py
```diff
@@ -63,3 +64,40 @@ def test_tar_output_encrypt(mock_fs: VirtualFilesystem, public_key: bytes, tmp_p
 
     with tarfile.open(name=decrypted_path, mode="r") as tar_file:
         assert entry.open().read() == tar_file.extractfile(entry_name).read()
+
+
+def test_tar_output_race_condition_with_growing_file(tmp_path: Path, public_key: bytes) -> None:
```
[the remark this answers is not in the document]
joost-j: Yes, added one for growing files as well

acquire/outputs/tar.py
```diff
@@ -100,7 +139,8 @@ def write(
             if stat:
                 info.mtime = stat.st_mtime
 
-        self.tar.addfile(info, fh)
+        with patch("tarfile.copyfileobj", copyfileobj):
```
Schamper, line 142: I'd probably rather we just inline the `addfile` logic.

Schamper (changes requested):

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+        self.tar.offset += len(buf)
+        bufsize = self.tar.copybufsize
+        if fh is not None:
+            bufsize = bufsize or 16 * 1024
+
+            if info.size == 0:
+                return
+            if info.size is None:
+                shutil.copyfileobj(fh, self.tar.fileobj, bufsize)
+                return
```
Schamper, line 125: I think we can remove this block since it would be an illegal action in this context.

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+
+            blocks, remainder = divmod(info.size, bufsize)
+            for _ in range(blocks):
+                # Prevents "long reads" because it reads at max bufsize bytes at a time
+                buf = fh.read(bufsize)
+                if len(buf) < bufsize:
+                    # raise exception("unexpected end of data")
```
Schamper, line 132:
```suggestion
```

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+                return
+            if info.size is None:
+                shutil.copyfileobj(fh, self.tar.fileobj, bufsize)
+                return
+
+            blocks, remainder = divmod(info.size, bufsize)
+            for _ in range(blocks):
+                # Prevents "long reads" because it reads at max bufsize bytes at a time
+                buf = fh.read(bufsize)
+                if len(buf) < bufsize:
```
Schamper, line 131: I think you can generalize this case instead of doing it twice. Keep track of how many bytes you actually wrote (i.e. using `.tell()` and only pad once.

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+                    # raise exception("unexpected end of data")
+                    # PATCH; instead of raising an exception, pad the data to the desired length
+                    buf += tarfile.NUL * (bufsize - len(buf))
+                self.tar.fileobj.write(buf)
+
+            if remainder != 0:
+                # Prevents "long reads" because it reads at max bufsize bytes at a time
+                buf = fh.read(remainder)
+                if len(buf) < remainder:
+                    # raise exception("unexpected end of data")
```
Schamper, line 141:
```suggestion
```

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+
+            if info.size == 0:
+                return
+            if info.size is None:
+                shutil.copyfileobj(fh, self.tar.fileobj, bufsize)
+                return
+
+            blocks, remainder = divmod(info.size, bufsize)
+            for _ in range(blocks):
+                # Prevents "long reads" because it reads at max bufsize bytes at a time
```
Schamper, line 129: Long or short?

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+                buf = fh.read(bufsize)
+                if len(buf) < bufsize:
+                    # raise exception("unexpected end of data")
+                    # PATCH; instead of raising an exception, pad the data to the desired length
+                    buf += tarfile.NUL * (bufsize - len(buf))
+                self.tar.fileobj.write(buf)
+
+            if remainder != 0:
+                # Prevents "long reads" because it reads at max bufsize bytes at a time
```
Schamper, line 138: Long or short?

acquire/outputs/tar.py
```diff
@@ -100,7 +102,54 @@ def write(
...
+        # Inline version of Python stdlib's tarfile.addfile & tarfile.copyfileobj,
+        # to allow for padding and more control over the tar file writing.
+        self.tar._check("awx")
+
+        if fh is None and info.isreg() and info.size != 0:
+            raise ValueError("fileobj not provided for non zero-size regular file")
+
+        info = copy.copy(info)
+
+        buf = info.tobuf(self.tar.format, self.tar.encoding, self.tar.errors)
```
Schamper, line 114: You could make this even safer by truncating to the previous offset/tar member end if any exception occurs while writing.

Schamper (changes requested):

acquire/outputs/tar.py
joost-j: So that means, when no size information is known we don't want to copy the file in the tar output? Because we do actually use that code right now in the stdlib.

joost-j: Ah I get it now, you write something to the tar file and return without adding the info to the `self.tar.members`. Wonder why this was implemented like this though. But will generalize it to do an early return when size is 0 or None.

acquire/outputs/tar.py
joost-j: Done, added a try/except around it to restore back to the original offset, plus a test case.

acquire/outputs/tar.py
joost-j: Don't know exactly what you mean by this, since it looks like we need to calculate the remainder for two different sizes (`bufsize` vs `tarfile.BLOCKSIZE`). But please do change / refactor the code if you can write this more concise.

acquire/outputs/tar.py
Schamper: I mean that I don't think `info.size` can be `None` because of the logic at the start of this method. Now we don't add it to the tar members. Is that desirable? I'm not sure what the behavior is now.

acquire/outputs/tar.py
```diff
@@ -100,7 +101,54 @@ def write(
...
+
+                blocks, remainder = divmod(tarinfo.size, tarfile.BLOCKSIZE)
+                if remainder > 0:
+                    self.tar.fileobj.write(tarfile.NUL * (tarfile.BLOCKSIZE - remainder))
+                    blocks += 1
+                self.tar.offset += blocks * tarfile.BLOCKSIZE
+
+            self.tar.members.append(tarinfo)
+        except Exception:
+            self.tar.fileobj.seek(saved_filepos)
```
Schamper, line 162: Log an error maybe?

acquire/outputs/tar.py
Schamper: I don't remember what I meant either.

tests/test_outputs_tar.py
```diff
@@ -63,3 +64,110 @@ def test_tar_output_encrypt(mock_fs: VirtualFilesystem, public_key: bytes, tmp_p
 
     with tarfile.open(name=decrypted_path, mode="r") as tar_file:
         assert entry.open().read() == tar_file.extractfile(entry_name).read()
+
+
+def test_tar_output_race_condition_with_shrinking_file(tmp_path: Path, public_key: bytes) -> None:
```
Schamper, line 70: Can you add a small docstring to these tests?

Schamper (changes requested):

acquire/outputs/tar.py
joost-j:
```
  """Copy length bytes from fileobj src to fileobj dst.
     If length is None, copy the entire content.
  """
```
    
Are the official docs. So we *should* actually do the `shutil.copyfileobj`, I'll just revert it to follow the stdlib.

PR #273, opened 2025-11-05
Add diagnostics

twiggler:
Add an option `--diagnostics <report.json>` to acquire which collects troubleshooting information about the host

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -32,6 +32,7 @@ from dissect.target.tools.utils import args_to_uri
 from dissect.util.stream import RunlistStream
 
 from acquire.collector import Collector, get_full_formatted_report, get_report_summary
+from acquire.diagnostics.common import diagnostics_info_json
 from acquire.dynamic.windows.named_objects import NamedObjectType
 from acquire.esxi import esxi_memory_context_manager
 from acquire.gui import GUI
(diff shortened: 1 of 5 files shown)


acquire/acquire.py
```diff
@@ -2399,6 +2400,30 @@ def main() -> None:
...
+            confirm = input("Do you want to continue? [y/N]: ").strip().lower()
+            if confirm not in ("y", "yes"):
+                print("Aborted diagnostics.")
+                exit_success(args.config.get("arguments"))
+            try:
+                # Determine output path
+                if isinstance(args.diagnostics, str):
+                    diag_path = Path(args.diagnostics)
+                elif log_file:
+                    diag_path = Path(log_file).with_name(Path(log_file).stem + "_diag.json")
```
Schamper: Isn't `log_file` already a `Path`?

acquire/diagnostics/linux.py
```diff
@@ -0,0 +1,232 @@
...
+import stat
+import subprocess
+from pathlib import Path
+
+
+def check_sudo() -> bool:
+    return os.geteuid() == 0
+
+
+def get_disk_info() -> dict:
```
Schamper: Adding suggestions for this all in GitHub is a bit slow, but can you add some newlines between sensible sections/parts?

acquire/utils.py
```diff
@@ -82,6 +82,16 @@ def create_argument_parser(profiles: dict, volatile: dict, modules: dict) -> arg
...
     output_group.add_argument("-of", "--output-file", type=Path, help="output filename")
 
+    output_group.add_argument(
+        "-dg",
+        "--diagnostics",
+        nargs="?",
+        const=True,
+        default=False,
+        metavar="[PATH]",
+        help="gather diagnostics information and print as JSON (optional path, defaults to logfile stem)",
```
Schamper:
```suggestion
        help="gather diagnostics information and save as JSON (optional path, defaults to logfile stem or diag.json if no logfile provided)",
```

Schamper (changes requested):
Is there some sanity unit test we can add? I.e. checking for errors, non empty values, that sort of thing. We run CI on Linux and Windows.

PR #274, opened 2025-11-05
Implicit set --compress when specifying --compress-method

Matthijsy:
When specifying `--compress-method` without `--compress` the first argument is ignored. This PR adds a quick check to see that you always specify `--compress` when you specify a method. There is no need to specify a method when the compression is not happing.

### DIFF
--- a/acquire/utils.py
+++ b/acquire/utils.py
@@ -371,6 +371,9 @@ def check_and_set_acquire_args(
             raise ValueError(
                 f"Invalid compression method for tar, allowed are: {', '.join(TAR_COMPRESSION_METHODS.keys())}"
             )
+    else:
+        if args.compress_method:
+            raise ValueError("--compress-method can only be used when --compress is set")
 
     if args.keychain_file:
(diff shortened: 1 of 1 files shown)


twiggler (dismissed):
The logic here correctly catches the edge case where a user supplies a method but forgets the boolean flag.

However, from a UX perspective, if the user explicitly supplies `--compress-method`, their intent is clearly to compress. Instead of raising a `ValueError`, would it be cleaner to simply infer `args.compress = True` if a method is present?

(I approved the pr in case you don't want to pursue this any further)

Matthijsy:
Good point, I agree that `--compress-method` should infer `--compress` as well. I just added that to the PR

twiggler (approved):

Matthijsy:
@twiggler Anything still needed to get this merged?

twiggler:
> @twiggler Anything still needed to get this merged?

@Matthijsy No, looks good.
The main branch is broken at the moment; after this has been addressed, I will merge it in (probably this week).

PR #280, opened 2026-01-08
Fix collection of `$Secure:$SII`

respondersGY:


acquire/acquire.py
```diff
@@ -366,11 +366,14 @@ def _run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector
...
                 "$MFT",
                 "$Boot",
-                "$Secure:$SII",
                 "$Secure:$SDS",
                 "$LogFile",
             ]
 
+            sii_fh = fs.ntfs.mft.get("$Secure").index("$SII")._index_stream
+
+            collector.output.write(fsutil.join(main_mountpoint, "$Secure:$SII"), sii_fh)
```
Miauwkeru: might be better to use `collector.write_bytes` assuming the stream isn't that large. Otherwise, maybe add a `collector.write` that writes the file header to it. E.g. something like this:

```python
class Collector:
    ...
    def write(self, destination_path: str, fh: BinaryIO) -> None:
        self.output.write(destination_path, fh)
        self.report.add_file_collected(self.bound_module_name, destination_path)
```
Then we can at least keep track that the file was collected. Besides that, the destination path would need to be joined with `collector.base` to be placed in `/fs/` properly.

acquire/acquire.py
```diff
@@ -366,11 +366,14 @@ def _run(cls, target: Target, cli_args: argparse.Namespace, collector: Collector
             filenames = [
                 "$MFT",
                 "$Boot",
-                "$Secure:$SII",
                 "$Secure:$SDS",
                 "$LogFile",
             ]
 
+            sii_fh = fs.ntfs.mft.get("$Secure").index("$SII")._index_stream
```
Miauwkeru: Instead of doing it like this, wouldn't the following work too?

`sii_fh = fs.ntfs.mft.get("$Secure").open("$SII", attr_type=...)`

Requires less knowledge about the internals of the internal ntfs structure.

Miauwkeru (commented):
One additional question, is `$Secure:$SII` inside the `$MFT` file itself and not on disk? Because if it is on disk, the method of getting `$Secure:$SII` will fail for those fs where `mountpoint` is `None`. E.g. unmounted ntfs filesystems

PR #293, opened 2026-03-19
Add --wait-exit option

Matthijsy:
This PR adds an argument called `--wait-exit` which requires the user to press any key before exiting. This is mainly useful when using acquire by double clicking an compiled version. Currently we have to open the log file to see if something went wrong, now we can inspect the logs before the terminal window closes.

### DIFF
--- a/acquire/acquire.py
+++ b/acquire/acquire.py
@@ -2334,19 +2334,29 @@ VOLATILE = {
 }
 
 
-def exit_success(default_args: list[str]) -> NoReturn:
+def exit_success(args: argparse.Namespace) -> NoReturn:
+    default_args = args.config.get("arguments")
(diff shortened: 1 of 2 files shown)


qmadev:
Correct me if I'm wrong, but `input()` requires the user to hit `Enter` before it returns right? Not any key?

Matthijsy:
Good point, you need to press enter. I updated the text

Matthijsy:
@qmadev Do you agree with the MR in this way? Anything needed to get this merged?

qmadev:
I'm not sure why you're asking me. I cannot merge anything.

Matthijsy:
Sorry, since we were in contact in this MR before, missed that you couldn't merge. Will wait for the maintainers in that case

PR #307, opened 2026-05-07
Add acquire version to report

### LINKED ISSUE #304: Add acquire version to Acquire report JSON output

qmadev:
closes #304


Schamper:
Are there currently tests for the reporting stuff? If so, could you add a test for this?

qmadev:
https://github.com/fox-it/acquire/blob/main/tests%2Ftest_collector.py#L40

This is the only test I could find searching for the keyword "report". So I think there are basically no tests for the reporting stuff.

Schamper (approved):

PR #311, opened 2026-08-03
Always use EsxiMemoryManager on ESXi

### LINKED ISSUE #310: Running `--list-children` breaks on ESXi
When running `acquire --list-children` on ESXi we get the following error:

```
 File "runpy", line 196, in _run_module_as_main
  File "runpy", line 86, in _run_code
  File "acquire_esxi.__main__", line 28, in <module>
  File "acquire.acquire", line 2357, in main
  File "acquire.utils", line 268, in parse_acquire_args
  File "dissect.target.tools.utils.cli", line 315, in list_children
  File "dissect.target.tools.utils.cli", line 256, in open_targets
  File "dissect.target.target", line 443, in open_all
dissect.target.exceptions.TargetError: Failed to find loader for local: path does not exist
```

After some digging I found that the issue is that the `EsxiMemoryManager` is not used when calling the `list_children` method, which makes other internals fail due to memory issues.

Matthijsy:
Currently the EsxiMemoryManager is only used for the acquisition part. However, this means that other code flows might fail (such as --list-children). This PR changes the behavior by always using the EsxiMemoryManager as long as we are on ESXi.

Fixes #310


acquire/acquire.py
```diff
@@ -2595,7 +2556,13 @@ def sort_files(files: list[str | Path]) -> list[Path]:
 
 if __name__ == "__main__":
     try:
-        main()
+        # Use esxi_memory_context_manager only if running on ESXi host
+        if platform.system().lower() == "vmkernel":
+            context_mgr = esxi_memory_context_manager()
+        else:
+            context_mgr = contextlib.nullcontext()
+        with context_mgr:
```
Schamper:
```suggestion
            ctx = esxi_memory_context_manager()
        else:
            ctx = contextlib.nullcontext()

        with ctx:
```
Just small cleanup from the original.

Schamper (changes requested):

Schamper (approved):

Schamper (changes requested):
Can you fix the linter?

Schamper (approved):

